PR #12, opened 2023-08-29
0032949: Modeling Algorithms - Infinite loop in ShapeUpgrade_UnifySameDomain

sebjf:
This PR addresses [0032949](https://tracker.dev.opencascade.org/view.php?id=32949): Infinite loop in ShapeUpgrade_UnifySameDomain.

As per the tracker, this occurs because the contour pcurve iteration in `RelocatePCurvesToNewUorigin` assumes that after the termination of the inner loop, `theVEmap` will be deprecated and eventually empty entirely, however it is possible to provide a shape for which this is not the case.

This change updates `RemoveEdgeFromMap()` and `RelocatePCurvesToNewUorigin()` to detect if this happens and terminate. 

It may be possible to avoid this condition by changing the logic behind the iteration, but detecting the case explicitly is actually quite trivial and has practically no overhead.

The issue can be reproduced by running UnifySameDomain on a shape. E.g. I just add this snippet to the start of one of the examples,

```
  TopoDS_Shape s;
  BinTools::Read(s, "D:\\repo\\ISSUE_633\\shape.brep");
  ShapeUpgrade_UnifySameDomain usd(s);
  usd.Build();
```

There is a file in the issue tracker to demonstrate it, but I have not been able to read it (BinTools doesn't seem to be aware of the format, even though the shape was only posted this year).

I also have a shape, but it is from a customer model so cannot be shared. If its not possible to reproduce with the shape from the issue tracker I can look again at getting permission.

### DIFF
--- a/src/ShapeUpgrade/ShapeUpgrade_UnifySameDomain.cxx
+++ b/src/ShapeUpgrade/ShapeUpgrade_UnifySameDomain.cxx
@@ -287,11 +287,12 @@ static Standard_Boolean TryMakeLine(const Handle(Geom2d_Curve)& thePCurve,
   return Standard_True;
 }
 
-static void RemoveEdgeFromMap(const TopoDS_Edge& theEdge,
+static Standard_Size RemoveEdgeFromMap(const TopoDS_Edge& theEdge,
                               TopTools_IndexedDataMapOfShapeListOfShape& theVEmap)
 {
   TopoDS_Vertex VV [2];
   TopExp::Vertices(theEdge, VV[0], VV[1]);
+  Standard_Size aNumRemovedEdges = 0;
   for (Standard_Integer i = 0; i < 2; i++)
   {
     if (!theVEmap.Contains (VV[i]))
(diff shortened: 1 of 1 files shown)


dpasukhi:
@sebjf Hello.
Thank you for your fix.
If you have not signed https://dev.opencascade.org/get_invoved/contributor_license_agreement
And don't want to sign it - i will merge your fix into main repo using my profile.
I will wait for your feedback.

sebjf:
Hi @dpasukhi, I have signed and submitted the CLA (though you can use your profile it that is easier!)

dpasukhi:
Great, i will transfer you commit from github to our repo on the next week (master for the current week is prepared)

luzpaz:
@dpasukhi was this done ? is there a x-post link ?

dpasukhi:
For now the fix is not correct. We need to continue research.

dpasukhi:
PR has branch in gitolite occt repo. See: https://tracker.dev.opencascade.org/view.php?id=32949
Closing for completing and testing fix by OCCT team.

PR #13, opened 2023-09-22
33487: Fix uninitialized structure StepVisual_FillAreaStyleColour for file with Unresoved Reference

Rachvel:
Fix of reported bug: [33487](https://tracker.dev.opencascade.org/view.php?id=33487)


luzpaz:
bump

dpasukhi:
My apologies for delay.
The patch is not fully correct. It is not right behavior to check object to null withing object.

We will check solution and fix with more common fix.

Rachvel:
Thank you for info.
I didn't had time to search the root of this exception. This workaround worked for us and I've reflected it in our local build.
At least you know that this situation can happened.

dpasukhi:
Original issue: [33487](https://tracker.dev.opencascade.org/view.php?id=33487)
Close PR because the fix will be on OCC side to proceed and test in our internal basics.

PR #15, opened 2024-02-26
0033749: Configuration - Update permission draw.sh

ThmsP:
Hello,
Currently, the file draw.sh is installed with very open permissions (777). 
This is probably not a problem for a user installation but we package it with rpm and thus it leaves a root owned file with very insecure permissions.
I have a very limited view of the project thus I may be missing the point of this permissions.

PR #22, opened 2024-06-27
0033752: Configuration - Update permission inspector.sh

dpasukhi:
Updated permission to  owner write only

Currently, the file inspector.sh is installed with very open permissions (777).
This is probably not a problem for a user installation but package it with rpm and thus it leaves a root owned file with very insecure permissions.


dpasukhi:
Issue: https://tracker.dev.opencascade.org/view.php?id=33752
Related PR: #15

dpasukhi:
Tests are success:
http://jenkins-test-10.nnov.opencascade.com/view/CR33752-master-dpasukhi/view/COMPARE/

ikochetkova (approved):
No remarks.

PR #24, opened 2024-06-28
0033615: Modeling Algorithms - Partition algorithm creates unexpected vertices

akaftase:
This problem occurs because of obtaining 2 ALines which share same vertex on the seam edge. There should be 2 different vertices with same(or almost the same) 3d parameters, but with different UV parameters because of periodic surface. Current fix allows to avoid the same vertices on seam edge and also checks the next vertex. Added consideration of predicted next point to avoid skipping tha point which is not on seam edge. Added test case for #33702

### DIFF
--- a/src/IntPatch/IntPatch_ALineToWLine.cxx
+++ b/src/IntPatch/IntPatch_ALineToWLine.cxx
@@ -647,6 +647,26 @@ void IntPatch_ALineToWLine::MakeWLine(const Handle(IntPatch_ALine)& theALine,
         {// Strictly equal!!!
           break;
         }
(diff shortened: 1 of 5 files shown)


akaftase:
Results of testing:
http://jenkins-test-10.nnov.opencascade.com/view/CR33615-master-akaftasev/view/COMPARE/

Note:
please, please integrate this branch together with [0033541](https://tracker.dev.opencascade.org/view.php?id=33541)

akaftase:
Please integrate this brach with #25

dpasukhi:
No remarks

dpasukhi (approved):

PR #61, opened 2024-09-09
Coding - Resolving C6263 warnings

dpasukhi:
Reorganizing code to not call alloca inside loop.
  alloca allocated memory from stack and free after finishing function


ikochetkova (approved):
No remarks.

PR #63, opened 2024-09-09
Coding - Resolving C6319 warning

dpasukhi:
Use of the comma-operator in a tested expression
  causes the left argument to be ignored when it has no side-effects.


ikochetkova (approved):
No remarks.

PR #64, opened 2024-09-10
Coding - Resolving C26439 & type formatting warnings

dpasukhi:
Sprintf  with %s always convert values into char*, not safety from int.
Move operators and constructors can be marked as noexcept


ikochetkova (approved):
No remarks.

PR #66, opened 2024-09-11
Coding - Precision.hxx file optimization

dpasukhi:
Precision.hxx optimized to have compiler-time
  constants for the most common floating-point values.
Reorganized code to avoid static jumping for parametric.


ikochetkova (approved):
No remarks.

PR #78, opened 2024-09-25
0027410: Data Exchange, Iges Import - Possible resource leak when parsing an invalid file

AtheneNoctuaPt:
fclose() call is added before early return statement in igesread() function.


dpasukhi (approved):
No remarks.
*Info for bugmaster
Needs to rename the comment to the new summary name

PR #89, opened 2024-10-07
Coding - Non-dependency include update

dpasukhi:
Removed or reorganized include that are not a part of the project dependency.
The new file for typedef is added because Select is not direct dependency.


ikochetkova (approved):
No remarks.

PR #94, opened 2024-10-11
Configuration - Implement VCPKG manifest basis

dpasukhi:
Implement the manifest for most part 3rd-parties.
Exclusions: tcl and tk.

Tested on Debian 12 with Clang 20.


dpasukhi:
Will be done with https://github.com/Open-Cascade-SAS/OCCT/pull/205

PR #106, opened 2024-10-19
Configuration - Production build type for linux

dpasukhi:
Fixed problem with not relevant flags and missed C language flags.


dpasukhi:
Interesting fact.
In the same machine STEP reading time
Windows (vc143) 670-750ms
Linux (clang20) 420-430ms

Just interesting point, can be useless ;)

PR #113, opened 2024-10-21
Tests - Building RefMan for 7.8

dpasukhi:


dpasukhi:
https://github.com/Open-Cascade-SAS/OCCT/actions/runs/11437409237/artifacts/2081499043
Artefact.

PR #130, opened 2024-10-29
Visualization - Physicalised Marker Scale, Line Rendering (Using a geometry shader)

winzlebee:
This patch introduces two (2) visualisation changes in the form of geometry shaders.

When point clouds are displayed in the OCCT viewer, sometime screen-space pixel values don't allow points to be easily identified. For our purposes, it was nice to have these occur as actual physical circular shapes with a configurable radius (marker size).

Line rendering in OpenCASCADE in the Core OpenGL profile suffers on some cards (MacOS for example) as the OpenGL implementation will not obey the line width and instead set a line width of 1 pixel. This patch also introduces a geometry shader which will be enabled for line rendering when geometry shading is supported by the OpenGL driver.

Physical scales for marker elements  are implemented in OpenGL using a geometry shader. This is currently exposed via the `SetMarkerPhysical` drawer attribute. I doubt this is the best way to express this - might be better to add this as an `Aspect3d_TypeOfMarker` or similar.

Some implementation notes;
- A geometry shader was chosen as it appears to fit in with pre-existing logic regarding mesh line presentation in (`Graphic3d_ShaderManager.cxx`). In the future, a mesh or instanced compute shader may be a better fit.
- Some new parameters needed to be provided to `Graphic3d_ShaderObject::CreateFromSource` such that differing numbers of geometry shader inputs and outputs can be specified
- The way this interacts with sprite markers (with textures) is untested, but will not work at the moment as the geometry shader does not export correct texture coordinates. This might either need to be made mutually exclusive or provided. In this case, a plane would be a more appropriate choice for generation too.
- The segments used for drawing the point polygon is hard-coded at the top of the file as `THE_NB_POINT_CIRCLE_SEGMENTS`
- The segments using for drawing line segment ends is hard-coded at the top of the file as `THE_LINE_END_CAP_SEGMENTS`

### DIFF
--- a/src/Draw/TKOpenGlTest/OpenGlTest/OpenGlTest_Commands.cxx
+++ b/src/Draw/TKOpenGlTest/OpenGlTest/OpenGlTest_Commands.cxx
@@ -182,6 +182,7 @@ void VUserDrawObj::Render(const occ::handle<OpenGl_Workspace>& theWorkspace) con
                                          Graphic3d_TypeOfShadingModel_Unlit,
                                          Graphic3d_AlphaMode_Opaque,
                                          false,
+                                         false,
                                          occ::handle<OpenGl_ShaderProgram>());
   aCtx->SetColor4fv(aColor);
 
--- a/src/Visualization/TKOpenGl/OpenGl/OpenGl_AspectsSprite.cxx
+++ b/src/Visualization/TKOpenGl/OpenGl/OpenGl_AspectsSprite.cxx
@@ -90,6 +90,7 @@ void OpenGl_AspectsSprite::UpdateRediness(const occ::handle<Graphic3d_Aspects>& 
   spriteKeys(theAspect->MarkerImage(),
              theAspect->MarkerType(),
              theAspect->MarkerScale(),
+             theAspect->MarkerPhysical(),
              theAspect->ColorRGBA(),
              aSpriteKeyNew,
              aSpriteAKeyNew);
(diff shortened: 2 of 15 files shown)


dpasukhi:
Thank you for you contribution!
I will assign reviewer soon (waiting for available viewer team member)
In case if you don't sign CLA, please proceed with https://dev.opencascade.org/get_involved
In case you already signed, please share ID from mail.
In case if you already signed a few hours ago, let me know, I will connect that request with you (CLA form a few hours ago were sent and now processing).

dpasukhi:
CLA signed: submission/523

gkv311:
@winzlebee 

Interesting patch. Just a couple of ideas.

Would be nice to have some screenshots attached to this MR to show how result actually looks. Thick lines emulated by geometry shader in Core Profile and conventional lines in Compatible Profile (with very thick values to see how it actually works). As well as spherical markers ;).

> This is currently exposed via the SetMarkerPhysical drawer attribute. I doubt this is the best way to express this - might be better to add this as an Aspect3d_TypeOfMarker or similar.

I think that indeed, it would be more natural extending `Aspect3d_TypeOfMarker` enumeration. I guess there is no idea behind to support marker pixmaps on these spheres?

> This patch also introduces a geometry shader which will be enabled for line rendering when geometry shading is supported by the OpenGL driver.

Wouldn't it be more clear to make a patch for lines and then make another patch for new feature for sphere markers on top of it?

winzlebee:
I've marked this PR as draft mode as I fix a few issues and re-base the commits to make more sense :)

> Would be nice to have some screenshots attached to this MR to show how result actually looks. Thick lines emulated by geometry shader in Core Profile and conventional lines in Compatible Profile (with very thick values to see how it actually works). As well as spherical markers ;).

I'll get those screenshots and post them soon!

> I think that indeed, it would be more natural extending `Aspect3d_TypeOfMarker` enumeration. I guess there is no idea behind to support marker pixmaps on these spheres?

I think it would be best if the spheres did support pixmaps, so I will implement this sometime in the coming days/weeks. This means that it would be possible to set any `Aspect3d_TypeOfMarker` but have the result appear as a screen aligned circle ("sphere"), in which case a separate drawer attribute might make more sense.

> Wouldn't it be more clear to make a patch for lines and then make another patch for new feature for sphere markers on top of it?

Yep. Probably I will split this PR into 3 separate ones, with a single commit each

1. Refactor geometry shader to emit the required input geometry count, output geometry count and input geometry type such that `Graphic3d_ShaderObject::CreateShader` has the required information to form the program 
2. Add a line renderer utilising a geometry shader, such that thick (>1px) lines can be rendered on all platforms
3. Implement geometry shader for circular markers that attenuates with viewer distance/zoom level

PR #155, opened 2024-11-12
Configuration - Optimize include copy logic

dpasukhi:
Avoiding re-copying the existed header file.
Add option to create a copy in case if symlink is not available.


AtheneNoctuaPt (approved):

dpasukhi:
Extra needs: check that type of includes inside build is the same type (symlink or direct include)

AtheneNoctuaPt (approved):

PR #156, opened 2024-11-12
Configuration - Option to extend debug symbols

dpasukhi:
New CMake variable created for updates compiler flags in debug mode.
Used only on GCC and Clang compilers.
As a result, the final product will higher size.
The changes is not final and can be reverted later, used for developers feedback.


dpasukhi:
No needs for changes.
That is working pretty well and redereferensing perfectly with clang or g++. No issues.
```
type": "cppdbg",
        "request": "launch",
        "name": "Launch",
        "MIMode": "gdb",
        "program": "${command:cmake.launchTargetPath}",
        "args": ["-v"],
        "runInTerminal": true,
        "setupCommands": [
            {
                "description": "Enable pretty-printing for gdb",
                "text": "-enable-pretty-printing",
                "ignoreFailures": true
            },
            {
                "description": "Set GDB to print objects",
                "text": "set print object on",
                "ignoreFailures": true
            },
            {
                "description": "Set GDB to print pretty",
                "text": "set print pretty on",
                "ignoreFailures": true
            }
        ],
```

PR #184, opened 2024-12-09
CR167: GLTF Import - Metadata support

AtheneNoctuaPt:
Limited support of metadata import is added to RWGltf_GltfJsonParser. Following Json data types are currently supported: int32, double, string, array of int32, array of double, array of strings, Json object. Notable unsupported types are: binary data, array of Json objects. Metadata is processed for nodes and meshes.
Tests "gltf_export" are updated with import testing and renamed to "gltf".

Method RWGltf_GltfJsonParser::gltfParseSceneNode() is slightly refactored: parsing of transformations and transformation matrices is moved into separate functions.


src/RWGltf/RWGltf_GltfJsonParser.hxx
```diff
@@ -297,17 +297,19 @@ protected:
   void bindNodeShape (TopoDS_Shape& theShape,
                       const TopLoc_Location& theLoc,
                       const TCollection_AsciiString& theNodeId,
-                      const RWGltf_JsonValue* theUserName)
+                      const RWGltf_JsonValue* theUserName,
```
dpasukhi, line 300: I would suggest to use `const Handle(TDataStd_NamedData)&`.
Originally input parameter must be not updated. It is originally a part of the new internal class.
The same for 2 more new methods in current file

src/RWGltf/RWGltf_GltfJsonParser.cxx
```diff
@@ -52,6 +53,245 @@ namespace
...
+  public:
+    //! Constructor. Initializes parser.
+    //! @param theParentID ID of the Json object that contains this "extras" value. Used only for printing messages.
+    //! @param theExtrasValue "extras" value to parse.
+    RWGltf_ExtrasParser(const TCollection_AsciiString& theParentID,
+                        const RWGltf_JsonValue& theExtrasValue);
+
+    //! Parses the "extras" value provided in the constructor.
+    //! @return Container with parsed data. May be nullptr if failed to parse.
+    Handle(TDataStd_NamedData) Parse();
```
dpasukhi, line 72: But the myResult allocated always. It is not related?
Can be it null? Or better to avoid allocate ahead?

tests/metadata/gltf/end
```diff
@@ -0,0 +1,77 @@
...
+} else {
+    if {$aSize != $ref_size} {
+        puts "Error: Wrong file size $aSize instead of $ref_size"
+    }
+	
+	if {$check_metadata == 1} {
+	  # Compare metadata in files.
+	  set aMetaDataLines [split ${aMetaData} "\n"]
+	  set aRefMetaDataLines [split ${ref_metadata} "\n"]
+	  set aMetaDataLinesCount [llength $aMetaDataLines]
```
dpasukhi: Please clear up the short tabs from here.
Imposters here :)

dpasukhi (approved):

PR #189, opened 2024-12-13
Shape Healing interface update

AtheneNoctuaPt:
New class XSAlgo_ShapeProcessor is added. It reimplements functionality of XSAlgo_AlgoContainer and makes it more convenient to use. XSAlgo_AlgoContainer is now internally uses methods of XSAlgo_ShapeProcessor when possible.
New overload of Perform() method is added to class ShapeProcess. It accepts a set of flags that defines required operations instead of string.

### DIFF
--- a/src/ShapeProcess/ShapeProcess.hxx
+++ b/src/ShapeProcess/ShapeProcess.hxx
@@ -16,11 +16,13 @@
 #ifndef _ShapeProcess_HeaderFile
 #define _ShapeProcess_HeaderFile
 
+#include <Message_ProgressRange.hxx>
 #include <Standard.hxx>
 #include <Standard_DefineAlloc.hxx>
 #include <Standard_Handle.hxx>
 
-#include <Message_ProgressRange.hxx>
+#include <bitset>
+#include <vector>
 
 class ShapeProcess_Operator;
 class ShapeProcess_Context;
@@ -34,10 +36,43 @@ class ShapeProcess_Context;
 class ShapeProcess 
 {
 public:
-
   DEFINE_STANDARD_ALLOC
 
+  //! Describes all available operations.
+  //! C++11 enum class is not used to allow implicit conversion to undelying type.
+  enum Operation : uint8_t
+  {
+    First = 0, // First operation index.
+    DirectFaces = First,
+    SameParameter,
+    SetTolerance,
+    SplitAngle,
+    BSplineRestriction,
+    ElementaryToRevolution,
+    SweptToElementary,
+    SurfaceToBSpline,
+    ToBezier,
+    SplitContinuity,
+    SplitClosedFaces,
+    FixWireGaps,
+    FixFaceSize,
+    DropSmallSolids,
+    DropSmallEdges,
+    FixShape,
+    SplitClosedEdges,
+    SplitCommonVertex,
+    Count // Number of operations.
+  };
+
+  // Bitset of operations. It is used to specify which operations should be performed.
+  // For example, to perform DirectFaces and SameParameter operations, use:
+  // ShapeProcess::OperationsFlags flags;
+  // flags.set(ShapeProcess::Operation::DirectFaces);
+  // flags.set(ShapeProcess::Operation::SameParameter);
+  // ShapeProcess::Perform(context, flags);
+  using OperationsFlags = std::bitset<Operation::Count>;
   
+public:
   //! Registers operator to make it visible for Performer
   Standard_EXPORT static Standard_Boolean RegisterOperator (const Standard_CString name, const Handle(ShapeProcess_Operator)& op);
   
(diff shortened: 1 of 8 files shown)


src/ShapeProcess/ShapeProcess.hxx
```diff
@@ -52,6 +87,28 @@ public:
...
+  //! Performs a specified sequence of operators on @p theContext.
+  //! @param theContext Context to perform operations on. Contains the shape to process
+  //!        and processing parameters. If processing parameters are not set, default values are used.
+  //!        Parameters should be in a scope of operation, for example,
+  //!        instead of "FromSTEP.FixShape.Tolerance3d"	we should use just "FixShape.Tolerance3d".
+  //! @param theOperations Bitset of operations to perform.
+  //! @param theProgress Progress indicator.
+  //! @return true if at least one operation was performed, false otherwise.
+  Standard_EXPORT static Standard_Boolean Perform(const Handle(ShapeProcess_Context)& theContext,
+                                                  const OperationsFlags               theOperations,
```
dpasukhi: Better to have as a reference *&bitset

src/XSAlgo/XSAlgo_ShapeProcessor.hxx
```diff
@@ -0,0 +1,138 @@
+// Created on: 2024-12-13
+// Created by: data exchange team
```
dpasukhi: No needs for these 2 lines

src/XSAlgo/XSAlgo_ShapeProcessor.hxx
```diff
@@ -0,0 +1,138 @@
...
+  //! @param theProgress Progress indicator.
+  //! @return Processed shape. May be the same as the input shape if no modifications were made.
+  Standard_EXPORT TopoDS_Shape ProcessShape(const TopoDS_Shape& theShape,
+                                            const OperationsFlags& theOperations,
+                                            const Message_ProgressRange& theProgress);
+
+  //! Get the context of the last processing.
+  //! Only valid after the ProcessShape() method was called.
+  //! @return Shape context.
+  Standard_EXPORT Handle(ShapeProcess_ShapeContext) GetContext() { return myContext; }
```
dpasukhi: No needs be EXPORT

src/XSAlgo/XSAlgo_ShapeProcessor.cxx
```diff
@@ -0,0 +1,548 @@
+// Created on: 2024-12-13
+// Created by: data exchange team
```
dpasukhi: No needs for these 2 lines

src/XSAlgo/XSAlgo_ShapeProcessor.cxx
```diff
@@ -0,0 +1,548 @@
...
+//=======================================================================
+//function : ToParameterMap
+//purpose  : 
+//=======================================================================
+XSAlgo_ShapeProcessor::ParameterMapPtr XSAlgo_ShapeProcessor::ToParameterMap(const DE_ShapeFixParameters& theParameters)
+{
+  // Helper lambda to convert enum to string.
+  auto makeString = [](const DE_ShapeFixParameters::FixMode theMode)
+    {
+      return std::to_string(static_cast<std::underlying_type<DE_ShapeFixParameters::FixMode>::type>(theMode));
```
dpasukhi: IS this magic c++11 support? Or we need to use macros?

src/XSAlgo/XSAlgo_ShapeProcessor.cxx
```diff
@@ -0,0 +1,548 @@
...
+//=======================================================================
+//function : XSAlgo_ShapeProcessor
+//purpose  : 
+//=======================================================================
+XSAlgo_ShapeProcessor::XSAlgo_ShapeProcessor(const ParameterMapPtr  theParameters,
+                                             const Standard_Real    theTolerance,
+                                             const Standard_Real    theMaxTolerance,
+                                             const TopAbs_ShapeEnum theDetalizationLevel,
+                                             const Standard_Boolean theNonManifold)
+  : myParameters(theParameters),
```
dpasukhi: In context if we have a MAP. I expect that not-initialized values that are not exists in map be filled up by default. Is it sounds reasonable? From struct DE_ShHealPar... Or probably we need additional methods with struct and map together.

dpasukhi (changes requested):
Thank you. Here a few remarks.
Solutions looks pretty well orginized.

src/XSAlgo/XSAlgo_ShapeProcessor.cxx
AtheneNoctuaPt: I think, everything used in this line of code is C++11 compatible. More specifically:
1. `std::underlying_type<T>` provides a member typedef type that names the underlying type of T. Note the I use here specifically `std::underlying_type<T>::type` ad not `std::underlying_type_t<T>`. The former was introduced in C++11, while latter was introduced in C++14.
2. `std::to_string()` - All overloads of this function were introduced in in C++11.

dpasukhi (approved):

PR #195, opened 2024-12-14
Testing - GH filter skipped tests

dpasukhi:
Add test skipping functionality and update test execution commands.
Skipped tests - tests where files are not public


dpasukhi:
[tests_to_skip.txt](https://github.com/user-attachments/files/18137106/tests_to_skip.txt)

dpasukhi:
We can't avoid 1400 tests. Moreover, the time is not reduced and the final archive size is similar

PR #201, opened 2024-12-18
Application Framework - Invalid document after partial read of XBF

dpasukhi:
Resolve missed refereces until all are resolved to produce expected
  result without exceptions.
XCAFDoc_ShapeTool::GetShape() throws Standard_NullObject exception after
  partial read of XBF document

XCAFDoc_ShapeTool::GetShape() throws Standard_NullObject exception when XBF document is read partially using functionality of PCDM_ReaderFilter.

The problem is that XBF driver resolves N+1 level of referenced items only while XCAFDoc_ShapeTool::GetShape() requires to follow the deeper chain of references.


dpasukhi:
Needs to refactor code and check careful with "while" scenario.
Additionally needs to move variables from the class field to method body if possible.

dpasukhi:
Original: https://tracker.dev.opencascade.org/view.php?id=33855

PR #202, opened 2024-12-18
Testing - Update GH Tests with parallel execution

dpasukhi:


dpasukhi:
Windows do not impacted. The problem on MESA and visualization performance

PR #211, opened 2024-12-26
Foundation Classes- Separate map's prime declaration

dpasukhi:
Refactor TCollection to use NCollection_Primes and remove obsolete TCollection.cxx
New prime declaration now will be more clear to use.


src/NCollection/NCollection_Primes.cxx
```diff
@@ -0,0 +1,60 @@
...
+                                      65003,    100019,   209953,   472393,    995329,    2359297,   4478977,    9437185,
+                                      17915905, 35831809, 71663617, 150994945, 301989889, 573308929, 1019215873, 2038431745};
+
+  // The number of prime numbers in the array.
+  constexpr int THE_NB_PRIMES      = sizeof(THE_PRIME_VECTOR) / sizeof(int);
+}
+
+int NCollection_Primes::NextPrimeForMap(const int theN)
+{
+  const int* aLow  = THE_PRIME_VECTOR;
```
AtheneNoctuaPt: Couldn't implementation just be like
```
auto aResult = std::lower_bound(std::begin(THE_PRIME_VECTOR), std::end(THE_PRIME_VECTOR), theN+1);
if (aResult == std::end(THE_PRIME_VECTOR))
{
  throw Standard_OutOfRange("NCollection_Primes::NextPrimeForMap() - requested too big size");
}
return *aResult;
```
Or performance isn't good enough in this case?

src/NCollection/NCollection_Primes.cxx
```diff
@@ -0,0 +1,60 @@
...
+  // size of array of buckets in the map.
+  // The prime numbers are used for array size with the hope that this will
+  // lead to less probability of having the same hash codes for
+  // different map items (note that all hash codes are modulo that size).
+  // The value of each next step is chosen to be ~2 times greater than previous.
+  // Though this could be thought as too much, actually the amount of
+  // memory overhead in that case is only ~15% as compared with total size of
+  // all auxiliary data structures (each map node takes ~24 bytes),
+  // and this proves to pay off in performance (see OCC13189).
+  constexpr int THE_PRIME_VECTOR[] = {101,      1009,     2003,     5003,      10007,     20011,     37003,      57037,
```
AtheneNoctuaPt: What's the advantage of using c-style array over std::array here?

src/NCollection/NCollection_Primes.cxx
dpasukhi: You code is better. Both of them doing binary search. (previously there was just for). It will be better then before.

src/NCollection/NCollection_Primes.cxx
dpasukhi: No advances. Move to array

dpasukhi:
https://github.com/Open-Cascade-SAS/OCCT/compare/cd67d304db2e53c66fd56d9986007928b0779abc..1b4bb62150655ac06be422fd16fc752f49bbba4f

Please check

AtheneNoctuaPt (approved):

PR #214, opened 2024-12-27
Coding - ASCII BRep double reading using stream

dpasukhi:
Simplify GetReal function by removing unnecessary buffer handling
_
Tempolary changes, if will be performance benefits - will rewrite on using stream directly on place


dpasukhi:
No benefits, no regressions.
As a result, we can move to stream directly. It can be done later ( as no benefits )

PR #215, opened 2024-12-27
Foundation Classes - Reorganize of OCCT RTTI

dpasukhi:
Implement functionality to avoid mutex operation and keeping unload data Making Standard_Type own specialized handle to avoid sharing and atomic operations

TODO:
Simplify the base type for type. Create extended type type. Simple will have ONLY reference to std::type_info and parent type. Extended will have a name and size of the object.


dpasukhi:
No needs to implement in 7.9

PR #226, opened 2024-12-29
Coding - Hiding symbols on unix

dpasukhi:
Enhance symbol visibility settings for GCC/Clang and update export macros


dpasukhi:
Can continue, but not so much benefits.
Close for now. Only a few megabytes are saved. Probably the most benefits for MacOS.

PR #237, opened 2025-01-04
Foundation Classes - Standard_Type remove type registry

dpasukhi:
Refactor Standard_Type registration by removing the registry and simplifying the Register method


dpasukhi:
GCC has regressions. Better to keep for now.

PR #246, opened 2025-01-06
Coding - Add clang-format configuration

dpasukhi:
```
New clang-format configuration added to determinate code style. The default version is 16.
Extend CMake to copy config file to build root.
Method/function separator deprecation:
  In case if function/method has declaration in header,
    definition must not have related comment.
    Only //==== [100 chars] ==== is allowed as a not connected
    separator.
  In case if function/method has  NOT declaration in header,
    definition must have related comment in doxygen style:
    // Descriptions
    // @param
    // @return
    Or just function/method separator:
    //==== [100 chars] ====
  All old separators with no description must be replaced to
    //==== [100 chars] ====
Header file comments:
  Comments in header files still started by //!
    Each public method must to have description in doxygen style. Each parameter must to be descripted.
Source file comments:
  Comments in source file always started by //
  Each internal classes/structures must to be descripted and each static method must to have at least short description in free form or doxygen style.

```
For example regex, that i was using to update method guards with no description in .cxx and .gxx files:
search:
`//==[=]+[\n\r]//[\s\n\r\t]*(?i:function)[\s\n\r\t]*:[\s\n\r\t]*[a-z0-9_:]*[\s\n\r\t]*//[\s\n\r\t]*(?i:purpose)[\s\n\r\t]*:[\s\n\r\t]*//===[=]+`
replace
`//=================================================================================================\n`

Current PR has 3 packages code style updated, chosen randomly to present new changes.
Updated packages(only cxx and hxx files):
- DE
- Approx
- ApproxInt

Used pwsh (powershell) command:
```
cd src/DE
Get-ChildItem -Recurse -Include *.hxx, *.cxx, *.lxx, *.gxx -Path ../DE, ../Approx, ../ApproxInt | ForEach-Object { clang-format -i -style=file:./../../.clang-format $_.FullName }
```
_
second commit regex targeted to single line separator with little details:
`//==[=]+[\n\r]+//[\s\n\r\t]*(?i:function)[\s\n\r\t]*:[\s\n\r\t]*[a-z0-9_:\t\s]*[\s\n\r\t]*//[\s\n\r\t]*(?i:purpose)[\s\n\r\t]*[a-z0-9_:\s\t.]*//===[=]+`
files: `src/DE/**, src/Approx/**, src/ApproxInt/**`
replace to `//=================================================================================================\n`
new regex: 
//-----[-]+
replace to:
`//=================================================================================================\n`


dpasukhi:
PR must be analyzed by each linear team internally.
Any comments must be sent by email or PR comment.
I will organize internal notification of more members to define single code style.

src/Approx/Approx_Curve2d.cxx
```diff
@@ -24,149 +23,173 @@
 #include <TColgp_Array1OfPnt2d.hxx>
 
 //=======================================================================
-//class : Approx_Curve2d_Eval
-//purpose: evaluator class for approximation
+// class : Approx_Curve2d_Eval
+// purpose: evaluator class for approximation
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_Curve3d.cxx
```diff
@@ -19,158 +19,173 @@
...
 #include <TColStd_HArray1OfReal.hxx>
+#include <TColgp_Array1OfPnt.hxx>
+#include <gp_Pnt.hxx>
+#include <gp_Vec.hxx>
 
 //=======================================================================
-//class : Approx_Curve3d_Eval
-//purpose: evaluator class for approximation
+// class : Approx_Curve3d_Eval
+// purpose: evaluator class for approximation
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurveOnSurface.cxx
```diff
@@ -24,82 +24,86 @@
...
+#include <TColgp_Array1OfPnt2d.hxx>
+#include <gp_Lin2d.hxx>
+#include <gp_Pnt.hxx>
+#include <gp_Vec.hxx>
 
 //=======================================================================
-//class : Approx_CurveOnSurface_Eval
-//purpose: evaluator class for approximation of both 2d and 3d curves
+// class : Approx_CurveOnSurface_Eval
+// purpose: evaluator class for approximation of both 2d and 3d curves
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurveOnSurface.cxx
```diff
@@ -109,254 +113,259 @@ void Approx_CurveOnSurface_Eval::Evaluate (Standard_Integer *Dimension,
...
+      break;
   }
 }
 
 //=======================================================================
-//class : Approx_CurveOnSurface_Eval3d
-//purpose: evaluator class for approximation of 3d curve
+// class : Approx_CurveOnSurface_Eval3d
+// purpose: evaluator class for approximation of 3d curve
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurveOnSurface.cxx
```diff
@@ -109,254 +113,259 @@ void Approx_CurveOnSurface_Eval::Evaluate (Standard_Integer *Dimension,
...
+      break;
   }
 }
 
 //=======================================================================
-//class : Approx_CurveOnSurface_Eval2d
-//purpose: evaluator class for approximation of 2d curve
+// class : Approx_CurveOnSurface_Eval2d
+// purpose: evaluator class for approximation of 2d curve
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurveOnSurface.cxx
```diff
@@ -109,254 +113,259 @@ void Approx_CurveOnSurface_Eval::Evaluate (Standard_Integer *Dimension,
...
+      break;
   }
 }
 
 //=============================================================================
-//function : Approx_CurveOnSurface
-//purpose  : Constructor
+// function : Approx_CurveOnSurface
+// purpose  : Constructor
 //=============================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//==========).

src/Approx/Approx_CurveOnSurface.cxx
```diff
@@ -109,254 +113,259 @@ void Approx_CurveOnSurface_Eval::Evaluate (Standard_Integer *Dimension,
...
+{
+  Perform(MaxSegments, MaxDegree, S, only3d, only2d);
+}
 
 //=============================================================================
-//function : Approx_CurveOnSurface
-//purpose  : Constructor
+// function : Approx_CurveOnSurface
+// purpose  : Constructor
 //=============================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//==========).

src/Approx/Approx_CurvilinearParameter.cxx
```diff
@@ -23,249 +23,275 @@
...
-Standard_IMPORT Standard_Real t_uparam;
+Standard_IMPORT Standard_Real    t_uparam;
 #endif
 
 //=======================================================================
-//class : Approx_CurvilinearParameter_EvalCurv
-//purpose  : case of a free 3D curve
+// class : Approx_CurvilinearParameter_EvalCurv
+// purpose  : case of a free 3D curve
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurvilinearParameter.cxx
```diff
@@ -23,249 +23,275 @@
...
+  std::cout << "number uparam calls = " << uparam_count << std::endl;
 #endif
 }
 
 //=======================================================================
-//class : Approx_CurvilinearParameter_EvalCurvOnSurf
-//purpose  : case of a curve on one surface
+// class : Approx_CurvilinearParameter_EvalCurvOnSurf
+// purpose  : case of a curve on one surface
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurvilinearParameter.cxx
```diff
@@ -276,159 +302,176 @@ Approx_CurvilinearParameter::Approx_CurvilinearParameter(const Handle(Adaptor2d_
...
+  std::cout << "number uparam calls = " << uparam_count << std::endl;
 #endif
 }
 
 //=======================================================================
-//function : Approx_CurvilinearParameter_EvalCurvOn2Surf
-//purpose  : case of a curve on two surfaces
+// function : Approx_CurvilinearParameter_EvalCurvOn2Surf
+// purpose  : case of a curve on two surfaces
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_CurvlinFunc.cxx
```diff
@@ -110,130 +118,142 @@ static void findfourpoints(const Standard_Real ,
...
 
-
 //=======================================================================
-//function : Init
-//purpose  : Init the values
-//history  : 23/10/1998 PMN : Cut at curve's discontinuities
+// function : Init
+// purpose  : Init the values
+// history  : 23/10/1998 PMN : Cut at curve's discontinuities
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//=======).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -17,63 +17,63 @@
...
 #include <Precision.hxx>
 #include <TColStd_Array1OfReal.hxx>
-#include <Geom2dAdaptor.hxx>
 
 //=======================================================================
-//class    : Approx_SameParameter_Evaluator
-//purpose  : Used in same parameterization curve approximation.
+// class    : Approx_SameParameter_Evaluator
+// purpose  : Used in same parameterization curve approximation.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -177,87 +172,93 @@ static Standard_Real ComputeTolReached(const Handle(Adaptor3d_Curve)& c3d,
...
   aDeviation = Max(aDeviation, Precision::Confusion()); // Tolerance in modeling space.
   return aDeviation;
 }
 
 //=======================================================================
-//function : Check
-//purpose  : Check current interpolation for validity.
+// function : Check
+// purpose  : Check current interpolation for validity.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -474,90 +476,100 @@ void Approx_SameParameter::Build(const Standard_Real Tolerance)
...
+
   myCurveOnSurface = Handle(Adaptor3d_CurveOnSurface)::DownCast(aData.myCOnS.ShallowCopy());
 }
 
 //=======================================================================
-//function : BuildInitialDistribution
-//purpose  : Sub-method in Build.
+// function : BuildInitialDistribution
+// purpose  : Sub-method in Build.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//=========).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -570,33 +582,36 @@ Standard_Boolean Approx_SameParameter::BuildInitialDistribution(Approx_SameParam
 }
 
 //=======================================================================
-//function : IncreaseInitialNbSamples
-//purpose  : Get number of C1 intervals and build new distribution on them.
+// function : IncreaseInitialNbSamples
+// purpose  : Get number of C1 intervals and build new distribution on them.
 //           Sub-method in BuildInitialDistribution.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//=========).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -620,26 +635,26 @@ Standard_Boolean Approx_SameParameter::IncreaseInitialNbSamples(Approx_SameParam
...
 
   return Standard_True;
 }
 
 //=======================================================================
-//function : CheckSameParameter
-//purpose  : Sub-method in Build.
+// function : CheckSameParameter
+// purpose  : Sub-method in Build.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//=======).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -744,12 +761,12 @@ Standard_Boolean Approx_SameParameter::CheckSameParameter(Approx_SameParameter_D
 }
 
 //=======================================================================
-//function : ComputeTangents
-//purpose  : Sub-method in Build.
+// function : ComputeTangents
+// purpose  : Sub-method in Build.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -781,42 +798,49 @@ Standard_Boolean Approx_SameParameter::ComputeTangents(const Adaptor3d_CurveOnSu
 }
 
 //=======================================================================
-//function : Interpolate
-//purpose  : Sub-method in Build.
+// function : Interpolate
+// purpose  : Sub-method in Build.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/Approx/Approx_SameParameter.cxx
```diff
@@ -825,71 +849,80 @@ Standard_Boolean Approx_SameParameter::Interpolate(const Approx_SameParameter_Da
 }
 
 //=======================================================================
-//function : IncreaseNbPoles
-//purpose  : Sub-method in Build.
+// function : IncreaseNbPoles
+// purpose  : Sub-method in Build.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/Approx/Approx_SweepApproximation.cxx
```diff
@@ -14,737 +14,843 @@
...
 #include <TColStd_Array1OfReal.hxx>
+#include <TColgp_Array1OfPnt2d.hxx>
+#include <gp_XYZ.hxx>
 
 //=======================================================================
-//class : Approx_SweepApproximation_Eval
-//purpose: evaluator class for approximation
+// class : Approx_SweepApproximation_Eval
+// purpose: evaluator class for approximation
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (Doxygen //! comments, I think?).

src/Approx/Approx_SweepApproximation.cxx
```diff
@@ -14,737 +14,843 @@
...
+                  Dichotomie);
+  }
 }
 
 //========================================================================
-//function : Approximation
-//purpose  : Call F(t) and store the results
+// function : Approximation
+// purpose  : Call F(t) and store the results
 //========================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/ApproxInt/ApproxInt_KnotTools.cxx
```diff
@@ -12,29 +12,29 @@
...
+// static const Standard_Real aSinCoeff = 0.30901699437494742410229341718282;
 static const Standard_Real aSinCoeff2 = 0.09549150281252627; // aSinCoeff^2 = (3. - Sqrt(5.)) / 8.
 static const Standard_Integer aMaxPntCoeff = 15;
 
 //=======================================================================
-//function : EvalCurv
-//purpose  : Evaluate curvature in dim-dimension point.
+// function : EvalCurv
+// purpose  : Evaluate curvature in dim-dimension point.
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/ApproxInt/ApproxInt_SvSurfaces.cxx
```diff
@@ -12,13 +12,10 @@
...
 
-
 #include <ApproxInt_SvSurfaces.hxx>
 
 //=======================================================================
-//function : ~ApproxInt_SvSurfaces
-//purpose  : Destructor
+// function : ~ApproxInt_SvSurfaces
+// purpose  : Destructor
 //=======================================================================
```
AtheneNoctuaPt: This comment should be updated to the new style (//===).

src/DE/DE_ShapeFixParameters.hxx
```diff
@@ -23,64 +23,64 @@ struct DE_ShapeFixParameters
...
+  FixMode          FixTailMode = FixMode::NotFix;
+  FixMode          MaxTailAngle = FixMode::NotFix;
+  FixMode          MaxTailWidth = FixMode::FixOrNot;
+  FixMode          FixSelfIntersectingEdgeMode = FixMode::FixOrNot;
+  FixMode          FixIntersectingEdgesMode = FixMode::FixOrNot;
   FixMode          FixNonAdjacentIntersectingEdgesMode = FixMode::FixOrNot;
-  FixMode          FixVertexPositionMode               = FixMode::NotFix;
-  FixMode          FixVertexToleranceMode              = FixMode::FixOrNot;
+  FixMode          FixVertexPositionMode = FixMode::NotFix;
+  FixMode          FixVertexToleranceMode = FixMode::FixOrNot;
```
AtheneNoctuaPt: Honestly, I think that previous version here (with AlignConsecutiveAssignments enabled, I think) looked a bit better. Just a bit easier to read. But that's probably just for me, so, any opinions?

AtheneNoctuaPt (commented):
Hello Dmitrii,
There are some remarks regarding deprecated functions/classes comments still remaining in the code. Since this update is supposed to represent the new style, we should probably fix that to avoid confusion.

src/Approx/Approx_Curve2d.cxx
dpasukhi: Updating cxx separators is not automated. I will update the regex to match more cases where information is useless. But in case if function has some more or less informative description, I would not want to lose it.
So, for now I will extend regex and add new regex to create new line between separator and function defining.

src/Approx/Approx_Curve3d.cxx
dpasukhi: Updating cxx separators is not automated. I will update the regex to match more cases where information is useless. But in case if function has some more or less informative description, I would not want to lose it.
So, for now I will extend regex and add new regex to create new line between separator and function defining.

src/Approx/Approx_CurveOnSurface.cxx
dpasukhi: Updating cxx separators is not automated. I will update the regex to match more cases where information is useless. But in case if function has some more or less informative description, I would not want to lose it.
So, for now I will extend regex and add new regex to create new line between separator and function defining.
_
//! used only for the information that will be transformed to PDF or HTML by doxygen and will be visible as a user manual.
//! will be useful only in .hxx files.

AtheneNoctuaPt (approved):

dpasukhi:
Can be cloned from the upstream repo for local checking:
https://github.com/Open-Cascade-SAS/OCCT/tree/clang-format

.clang-format
vttrtp, line 1: as proposial can be `BreakConstructorInitializers: BeforeColon`:
image
`BreakInheritanceList: BeforeColon`
image
`BreakBeforeTernaryOperators: true`
image
`BreakBeforeBinaryOperators: NonAssignment`

image
the provided changes can be useful in long expressions (sometimes I split screen, and right part of page can be hidden, but symbols like :,&&,? will be on begin  of row, which allow to don't do additional scroll, to read them). Another argument when we try to comment the row to make some checks - in this case we need to remove symbol from another row which can be more time consuming and not convenient. So from my point of view these changes are more representative and usable

.clang-format
```diff
@@ -0,0 +1,38 @@
...
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BreakConstructorInitializers: AfterColon
+ColumnLimit: 100
+ContinuationIndentWidth: 2
+IndentCaseLabels: true
+IndentWidth: 2
+IndentWrappedFunctionNames: true
+PackConstructorInitializers: Never
```
vttrtp, line 29: maybe there is sense to remove this parameter from config  (default PackConstructorInitializers: CurrentLine)to have short constructors such as:
```
Constructor() : a(), b()

```

.clang-format
vttrtp: maybe also `SpaceAfterTemplateKeyword: false`
to don't produce extra space:
```
template <typename T> -> template<typename T> (stl style)
```

dpasukhi: Here is my opinion and it is not final. I would be glad if someone else will be able to share their view.
_
`BreakConstructorInitializers: AfterColon` used to make all arguments in the same condition to fast reorder and manipulate. If we will have : in the same line it will give special attention to first field. And based on some `#ifdef` cases fields can be initialize differently. So, I would recommend to have all initializations to have the same logic and style.
From my side I would recommend: `AfterColon `
_
`BreakInheritanceList: BeforeColon` I would avoid with the same idea as first option related to have the same attention to any inheritance classes. What about choose `AfterColon`? It make all parents in the same level and helps to clear read the logic. BUT OCCT RTTI mostly avoid multiple inheritance and we usually have only 1 parent. Based on it we can use `AfterComma`.
From my side I would recommend: `AfterColon `
_
`BreakBeforeBinaryOperators: NonAssignment`. Yes, in case of git diff or splitting screen it can helps to avoid some extra actions to read code. But in the same time it will leads to difficulties to understand the operation type because the operators are randomly located. The current colon limitation is 100 symbols. For sure it is not comfortable for splitting, but still enough for general code). Usually so complex operations are short in line.
So, benefit - clear view on tiny screen.
Misadventures - less comfort to read code in normal development time.
As for a code commenting - IDE helps to comment selected peace of code with /**/. In that case developer truly will be sure which peace of operation he will comment. (commenting || or && can leads to different results).
From my side I would recommend: `None`

.clang-format
dpasukhi: Making inline initialization list leads to possible issue with incorrect ordering and complicity with reorder. Making attentions on initialization order is very important in C++ because we have depends on class packing and location in memory.
From my side i would keep the current style.

.clang-format
vttrtp: ```IndentPPDirectives: BeforeHash```
will look more readable:
image

dpasukhi:
```
maybe also SpaceAfterTemplateKeyword: false
to don't produce extra space:
```
Yes, I agree. If someone have another view, welcome.

.clang-format
AtheneNoctuaPt: I agree here with Dmitrii. For me personally it is somewhat easier to recognize correct/incorrect order when initializers are always ordered in the same way as respective fields are declared in class - from top to down. Besides, while the example with just `a(), b()` looks pretty, realistically majority of classes in our codebase would look like:
```
Constructor(int theArgument) : myFirstField(), myData(theArgument), myOtherField(), myFlag(true)
```
which is harder to read, in my opinion, compared to
```
Constructor(int theArgument)
: myFirstField(),
  myData(theArgument),
  myOtherField(),
  myFlag(true)
```

.clang-format
AtheneNoctuaPt: 1. `BreakConstructorInitializers` - `AfterColon` makes it easier to reorder, as was mentioned. I prefer that.

2. `BreakInheritanceList` - I would prefer `AfterComma`, I guess. Since we (fortunately) do not use multiple inheritance that often. But, honestly, it doesn't matter for me a lot, so I would go with whatever you guys like more.

3. `BreakBeforeTernaryOperators` - `true` looks good. I think it would be easier to notice during quick code reading.

4. `BreakBeforeBinaryOperators` - Okay, so maybe we should check how it actually looks. I tested `NoneAssignment` and   for me it looks like this:
  ```
      double aBcd = smth.getPosition().x + smthRadius - rightPaddle.getPosition().x
                  - paddleSize.x / 2 + smth.getPosition().x + smthRadius
                  - rightPaddle.getPosition().x + smth.getPosition().y + smthRadius
                  - rightPaddle.getPosition().y - paddleSize.y / 2 + smth.getPosition().y
                  - smthRadius - rightPaddle.getPosition().y + paddleSize.y / 2;
  ```
  Maybe it is defined by different parameters but for me it was aligned along '=' operator. I like it, I think it looks quite readable. Well, as readable as this artificial over-complicated example can look.

5. `SpaceAfterTemplateKeyword: false` - I agree, let's follow STL style.

6. `IndentPPDirectives: BeforeHash` - Makes sense. For small macros it would probably be more readable. And when code (like .cpp file, for example) is polluted with macros for conditional compilation, I guess that indentation of the preprocessor directives that is currently on your screen would help to understand how deep you are inside those directives.

.clang-format
vttrtp: Actually in microsoft version of stl they use spaces after template keyword but in libstd no 
image
image

dpasukhi: GNU/Microsoft coding style. We are using Microsoft by default.
So, only fields that on .clang-format they are creating differences from the some common styles.
The main attention in modification are for packing, sorting and splitting. Spaces better to keep as in default style. What do you think?

AtheneNoctuaPt: > GNU/Microsoft coding style. We are using Microsoft by default. So, only fields that on .clang-format they are creating differences from the some common styles. The main attention in modification are for packing, sorting and splitting. Spaces better to keep as in default style. What do you think?

For me, any will do.

.clang-format
martin-bernhard: For points 1) and 2) I personally prefer and use the `BeforeComma` style. It allows me to easily comment out a single line or reorder them (except of the first with the colon).

vttrtp: @dpasukhi what about `BreakBeforeTernaryOperators - true` ?

vttrtp: @martin-bernhard I like it more too than after colon

dpasukhi: `BeforeComma` in context of (1) and (2) are making attention for the developer on the some not useful syntax.
In that context `:` or `,` has no so much impact on the clarity of the code.
What I mean - that symbol usually ignored by developer when looking into code. I guess better to have attension on fields itself then have "symbol_space_field".
In current version we have clear view on the fields and their position then on some syntax. All syntax stuff put on the out-of-attention position (first line and end of the current line).
From my side I would recommend to stay with `AfterColon`.
(2) `AfterColon  `equal with `AfterComma` in our context in case if class have enough space to locate. So for average cases we will have no see differences, but in rare cases `AfterComma` will have more clarity.

martin-bernhard: I just wrote a short script to search on Github what others are using in `.clang-format`. As the API only returns a maximum number of 100 search results, I only created a statistic for that number. Not all found `.clang-format` files had the specific options set:

**BreakInheritanceList Statistics:**  
|Option|Usage|
|:---:|:---:|
|BeforeColon|64.3%|
|AfterColon|15.5%|
|BeforeComma|17.9%|
|AfterComma|2.4%|

**BreakConstructorInitializers Statistics:**  
|Option|Usage|
|:---:|:---:|
|BeforeColon|56.2%|
|BeforeComma|27.0%|
|AfterColon|16.9%|

vttrtp: Additional argument for BeforeColon that dots on the beginning help you to understand that in this place we start to define superclasses or parameters and nothing above

dpasukhi: The GitHub research about the usage of some parameters in clang-format doesn't have a direct equivalent for the "comfort" style in our project.
The idea of .clang-format is to use the default style and then update some project-related options that have specific benefits for their team, their product, and its lifecycle.
I checked the GNU and Microsoft coding styles, and they do indeed use BeforeColon.
In that case, I would **agree** to use it, based on common practice.
I agree with that approach and will remove the override to use the default Microsoft style.
_
Additionally, let me share the main purpose of the new code style formatting:
Most GitHub C++ projects are small pet-projects. OCCT, however, is a huge codebase with code that hasn't been modified by humans for decades.
But that code is still read by humans from time to time. That is why the main direction for the new formatting style is clarity for reading and review.

dpasukhi: BeforeColon will have extra 2 spaces at the beginning. But oke, if 2 of you and microsoft have that I agree with that price.
You can check last modifications.

dpasukhi:
If no more suggestions or recommendations - will be merge today evening into IR.
19.01 will be merged to "master"

PR #256, opened 2025-01-14
Testing - WebAssembly build validation

dpasukhi:
Add GitHub Actions workflow for WebAssembly build validation on Windows using MinGW


dpasukhi:
Some issues with vcpkg. Will investigate

AtheneNoctuaPt (approved):

PR #260, opened 2025-01-16
Coding - Separate DE Wrapper's plugins to own package

dpasukhi:
Reorganize DE Wrapper classes to have single style and logic.
Each TKDE* will have own DE*_Provider and DE*_ConfigurationNode
DE* package will have all necessary classes and enums inside instead of another places.


dpasukhi:
The comment include only moving ConfigurationNodes and Providers to separated package and marking previous one as deprecated.
And formatting of the moved classes. With no any changes.

AtheneNoctuaPt (approved):

Issue #278, opened 2025-01-22
Fail to export to GLTF with draco buffer index out of bounds


elias-bananaz:
### Description

Reading a Step file, and saving it to a Gltf with Draco compression (and MergeFaces) results in a crash for some files.

### Expected Behavior

Model is saved as GLTF.

### Actual Behavior

Exception is thrown and the model is not saved.

Exception happens in [RWGltf_CafWriter.cxx](https://github.com/Open-Cascade-SAS/OCCT/blob/c24871f654975e121cf536230ac57f4d72a49731/src/RWGltf/RWGltf_CafWriter.cxx#L1822):
```
myWriter->Int(myBuffViewsDraco[theDracoBufInd].Id);
```

![Image](https://github.com/user-attachments/assets/7608051a-02fc-4ba1-b4f4-7f56a126bab7)

### Sample Code or DRAW Tcl Script

```
#include <iostream>
#include <filesystem>
#include <STEPCAFControl_Reader.hxx>
#include <XCAFApp_Application.hxx>
#include <BinXCAFDrivers.hxx>
#include <TDocStd_Document.hxx>
#include <XCAFDoc_ShapeTool.hxx>
#include <RWGltf_CafWriter.hxx>
#include <BRepMesh_IncrementalMesh.hxx>
#include <TopTools_HSequenceOfShape.hxx>
#include <XCAFDoc_DocumentTool.hxx>

void ConvertSTEPtoGLTF(std::string path)
{
    Handle(TDocStd_Application) app = new TDocStd_Application();
    BinXCAFDrivers::DefineFormat(app);
    Handle(TDocStd_Document) doc;
    app->NewDocument("BinXCAF", doc);

    STEPCAFControl_Reader reader;
    IFSelect_ReturnStatus status = reader.ReadFile(path.c_str());
    if (status != IFSelect_RetDone)
    {
        return;
    }

    reader.Transfer(doc);

    Handle(XCAFDoc_ShapeTool) shapeTool = XCAFDoc_DocumentTool::ShapeTool(doc->Main());
    TDF_LabelSequence labels;
    shapeTool->GetFreeShapes(labels);

    for (Standard_Integer i = 1; i <= labels.Length(); ++i)
    {
        TopoDS_Shape shape = shapeTool->GetShape(labels.Value(i));
        if (!shape.IsNull())
        {
            BRepMesh_IncrementalMesh mesh(shape, 0.1, false, 0.5, true);
        }
    }

[...]

PR #279, opened 2025-01-22
Coding - Fix dracobuffer index out of bounds

elias-bananaz:
## Description
Fixes an exception where the bufferIndex is out of bounds. Happens when Draco compression is combined with `SetMergeFaces`.
There seems to be a mismatch between the amount of buffers created and the amount of gltf faces.

## The problem
Before calling `writePrimArray (*aGltfFace, aNodeName, aDracoBufInd, toStartPrims)` [here](https://github.com/Open-Cascade-SAS/OCCT/blob/f91655b0c5de3c4cd49d4042959dd6c7def0fa74/src/RWGltf/RWGltf_CafWriter.cxx#L1910) there's a check to get the index from `aDracoBufIndMap` if it exists, or add a new one if it doesn't. The index is then incremented.

This is done for each `RWGltf_GltfFace` in  `aGltfFaceList`.

The problem is that in the `writeBinData` function, the Draco compressed meshes `aMeshes` are added once for all `aGltfFaceList`. This probably worked as long as aGltfFaceList's length is 1, but otherwise, there should be a mesh created for each `RWGltf_GltfFace`.

I also simplified the logic before calling `writePrimArray` to make it more readable and easy to understand what's happening.

Fixes issue https://github.com/Open-Cascade-SAS/OCCT/issues/278

### DIFF
--- a/src/RWGltf/RWGltf_CafWriter.cxx
+++ b/src/RWGltf/RWGltf_CafWriter.cxx
@@ -746,26 +746,26 @@ bool RWGltf_CafWriter::writeBinData (const Handle(TDocStd_Document)& theDocument
         continue;
       }
       
-      std::shared_ptr<RWGltf_CafWriter::Mesh> aMeshPtr;
-#ifdef HAVE_DRACO
-      ++aMeshIndex;
-      if (myDracoParameters.DracoCompression)
-      {
-        if (aMeshIndex <= aMeshes.size())
-        {
-          aMeshPtr = aMeshes.at(aMeshIndex - 1);
-        }
-        else
-        {
-          aMeshes.push_back(std::make_shared<RWGltf_CafWriter::Mesh>(RWGltf_CafWriter::Mesh()));
-          aMeshPtr = aMeshes.back();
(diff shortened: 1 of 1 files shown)


dpasukhi:
Dear @elias-bananaz could you please share you Contributor ID if you already sign one.
If not - please follow the instruction:
https://dev.opencascade.org/get_involved/cla_processing_rules
*_All fields that are not mark as optional must to be full filled._

elias-bananaz:
> Dear @elias-bananaz could you please share you Contributor ID if you already sign one. If not - please follow the instruction: https://dev.opencascade.org/get_involved/cla_processing_rules *_All fields that are not mark as optional must to be full filled._

CLA id: 1118

dpasukhi:
@elias-bananaz my apologies. I found the CLA.
Thank you for you help!
Now we are moving to test your changes on our Test system, then move to review.
The time line for internal processes - current week.

dpasukhi:
No regressions.

dpasukhi (approved):

PR #284, opened 2025-01-24
Data Exchange - Losing attributes on NonManifold STEP

dpasukhi:
Move step related processing from XSBase to STP module.
Added history apply after reading of NonManifold


AtheneNoctuaPt (approved):

dpasukhi:
No regressions

PR #307, opened 2025-01-30
Step thread safety improvement

AtheneNoctuaPt:
Mutex is added to XSControl_WorkSession to prevent data races during reading and writing.

### DIFF
--- a/src/XSControl/XSControl_WorkSession.hxx
+++ b/src/XSControl/XSControl_WorkSession.hxx
@@ -196,6 +196,7 @@ private:
   //! Clears binders
   Standard_EXPORT void ClearBinders();
 
+private:
   Handle(XSControl_Controller)     myController;
   Handle(XSControl_TransferReader) myTransferReader;
   Handle(XSControl_TransferWriter) myTransferWriter;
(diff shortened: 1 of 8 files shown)


src/XSControl/XSControl_WorkSession.hxx
```diff
@@ -196,11 +196,14 @@ private:
...
   Standard_EXPORT void ClearBinders();
 
+private:
   Handle(XSControl_Controller)     myController;
   Handle(XSControl_TransferReader) myTransferReader;
   Handle(XSControl_TransferWriter) myTransferWriter;
   XSControl_WorkSessionMap         myContext;
   Handle(XSControl_Vars)           myVars;
+
+  static Standard_Mutex myGlobalMutex; //!< Mutex to prevent data races during reading and writing.
```
dpasukhi: Please hide it into singleton inside CXX

dpasukhi (changes requested):
Here a few remarks and please add some tests with public CAD files to validate your new commands.

dpasukhi (approved):

PR #363, opened 2025-02-12
Data Exchange - STEP Dimensions's plane export without presentation

AtheneNoctuaPt:
Dimension plane is now correctly exported even with absent tesselation.


dpasukhi (approved):

dpasukhi:
No needs for patch

Issue #356, opened 2025-02-10
Double click not detected after long click


Hyarius:
### Description

Currently, in the AIS_ViewController, we have the method UpdateMouseButtons (AIS/AIS_ViewController.cxx:712)

Inside it, the library increment the index by one each time we trigger a mouse button, we increment a counter by one. And this counter is only reset when the mouse click is detected as a double click. Which happen only if the counter is 2 and the timer is not elapsed.

Now, we have a problem : when the user click on the view, don't release the button for a long period of time (~200ms), don't move the mouse, than release the button, we cannot receive any double click anymore : the counter has been raised to higher than 2, and therefore cannot go back to 0 as we don't detect any double clicks.

We propose to change this by editing the function as proposed in the attached files.

### Expected Behavior

1) Press any mouse button received by the AIS_ViewController on the view
2) Don't release the button, do not move it, for longer than the duration of the double click
3) Once this duration is passed, release the mouse button
4) Try to double click - It should be detected

### Actual Behavior

1) Press any mouse button received by the AIS_ViewController on the view
2) Don't release the button, do not move it, for longer than the duration of the double click
3) Once this duration is passed, release the mouse button
4) Try to double click - It shouldn't be detected

This behavior can be "fixed" by moving the camera (rotate or translation), as it will reset the counter to 0, relaunching the double click check

### Sample Code or DRAW Tcl Script

Any program with a AIS_ViewController should have the problem.

### Operating System

Windows

### Compiler

Clang

### Bitness

64-bit

### OCCT Version

latest

HyariusOld:
I propose to change the content of the first if with this :

```cpp
if (double(aDelta.cwiseAbs().maxComp()) < aTolClick)
{
	++myMouseClickCounter;

	const bool isCounterValid = myMouseClickCounter == 2;
	const bool isTimerStarted = myMouseClickTimer.IsStarted();
	const bool isTimerElapsed = myMouseClickTimer.ElapsedTime() > myMouseDoubleClickInt;

	const bool isTimerValid = isTimerStarted && !isTimerElapsed;

	const bool isDoubleClick = isCounterValid && isTimerValid;

	// Let's take this event timeline :
	//    A          B   C      
	// ---o----------o---o------->

	// When event A occured, the timer is started, and the counter is raised to 1
	// When event B occured, later on, the timer is elapsed, and the counter have been raised to 2.
	// As the timer isn't valid anymore, the double click was not detected as true, and therefore the
[...]

PR #385, opened 2025-02-20
Refactor mouse click handling logic for improved double-click detection

jboissy-mediasofts:
Let's take this event timeline :
```
   A          B   C      
---o----------o---o------->
```

When event A occured, the timer is started, and the counter is raised to 1
When event B occured, later on, the timer is elapsed, and the counter have been raised to 2.
As the timer isn't valid anymore, the double click was not detected as true, and therefore the
counter isn't reset.
The timer, on the other hand, is reset as needed.
When event C occured, the counter is increased to 3, which will  deny us from detecting the
double click anymore.
Note : as we reset the counter to 0 in other function (in mouse update position, mainly) we
can work around this issue by sending mouse motion to the application.

We should be able to "detect" that the event B isn't a valid double click, and try to use it as
the "first" click of a futur double click.
To do this, we propose to check if the timer is valid, and if not, set the counter to one, meaning
it is suppose to be the first click.

Pull request to fix the problem explain in the following ticket :
https://github.com/Open-Cascade-SAS/OCCT/issues/356

### DIFF
--- a/src/AIS/AIS_ViewController.cxx
+++ b/src/AIS/AIS_ViewController.cxx
@@ -695,8 +695,21 @@ bool AIS_ViewController::UpdateMouseButtons(const Graphic3d_Vec2i& thePoint,
     if (double(aDelta.cwiseAbs().maxComp()) < aTolClick)
     {
       ++myMouseClickCounter;
-      const bool isDoubleClick = myMouseClickCounter == 2 && myMouseClickTimer.IsStarted()
-                                 && myMouseClickTimer.ElapsedTime() <= myMouseDoubleClickInt;
+
+      const bool isCounterValid = myMouseClickCounter == 2;
+      const bool isTimerStarted = myMouseClickTimer.IsStarted();
+      const bool isTimerElapsed = myMouseClickTimer.ElapsedTime() > myMouseDoubleClickInt;
+
+      const bool isTimerValid = isTimerStarted && !isTimerElapsed;
+
+      const bool isDoubleClick = isCounterValid && isTimerValid;
+
+      if (!isTimerValid)
+      {
+        myMouseClickCounter = 1;
+      }
+
+      myMouseClickCounter %= 2;
 
       myMouseClickTimer.Stop();
       myMouseClickTimer.Reset();


dpasukhi:
Dear @jboissy-mediasofts 
Thank you for your patch!

Contributing to OCCT require signing of CLA (more details [Contributing.md](https://github.com/Open-Cascade-SAS/OCCT/blob/858dd02fed80c8ca68cf1d70a6e03ecf97f849ca/.github/CONTRIBUTING.md) or [Get Involved](https://dev.opencascade.org/get_involved) )

If you are already have signed CLA, please share the ID that we shared with your by email.

Best regards, Dmitrii.

src/AIS/AIS_ViewController.cxx
```diff
@@ -695,8 +695,21 @@ bool AIS_ViewController::UpdateMouseButtons(const Graphic3d_Vec2i& thePoint,
...
+      const bool isTimerValid = isTimerStarted && !isTimerElapsed;
+
+      const bool isDoubleClick = isCounterValid && isTimerValid;
+
+      if (!isTimerValid)
+      {
+        myMouseClickCounter = 1;
+      }
+
+      myMouseClickCounter %= 2;
```
mzernova, line 712: It's better to keep only this line
It will work and simplify the logic

dpasukhi: It is a final suggestion? It can be 0-1 only as a result always. Need to protect to have at least one click, not?
For sure multiple definitions are complex,but it is readable and looks complete.
Please check that that solution with %2 is enough, I'm not fully sure.

src/AIS/AIS_ViewController.cxx
```diff
@@ -695,8 +695,21 @@ bool AIS_ViewController::UpdateMouseButtons(const Graphic3d_Vec2i& thePoint,
     if (double(aDelta.cwiseAbs().maxComp()) < aTolClick)
     {
       ++myMouseClickCounter;
-      const bool isDoubleClick = myMouseClickCounter == 2 && myMouseClickTimer.IsStarted()
-                                 && myMouseClickTimer.ElapsedTime() <= myMouseDoubleClickInt;
```
mzernova, line 699: And return this check

src/AIS/AIS_ViewController.cxx
EreliaStudio: The solution with only the line 712 will NOT fix the problem.

jboissy-mediasofts: The solution with only the line 712 will NOT fix the problem. 


As i said in the issue :
```
// Let's take this event timeline :
//    A          B   C      
// ---o----------o---o------->

// When event A occured, the timer is started, and the counter is raised to 1
// When event B occured, later on, the timer is elapsed, and the counter have been raised to 2.
// As the timer isn't valid anymore, the double click was not detected as true, and therefore the
// counter isn't reset.
// The timer, on the other hand, is reset as needed.
// When event C occured, the counter is increased to 3, which will  deny us from detecting the
// double click anymore.
// Note : as we reset the counter to 0 in other function (in mouse update position, mainly) we
// can work around this issue by sending mouse motion to the application.

// We should be able to "detect" that the event B isn't a valid double click, and try to use it as
// the "first" click of a futur double click.
// To do this, we propose to check if the timer is valid, and if not, set the counter to one, meaning
// it is suppose to be the first click.
```

Not that with your proposition (Only keeping the %=2), at event C, the counter will be 1, not 3, but the problem will still remain as it isn't 2

dpasukhi: Original suggestion was from mzernova:
https://github.com/Open-Cascade-SAS/OCCT/pull/385

In case if indeed the suggestion is not relevant - then I have no remarks for the current fix.
@mzernova do you have remarks on the PR's code?

mzernova (approved):

PR #429, opened 2025-03-12
Configuration - Resource structure reorganization

dpasukhi:


dpasukhi:
TODO: update env variables, direct paths to the resources and installation

PR #431, opened 2025-03-12
0033607: Visualization - Implementation of hardware occlusion queries

Hossam86:
Hardware occlusion queries provide a mechanism to determine whether any pixels would be drawn. For instance, it can be used as a visibility filter any pixels would be drawn. for instance, it can be used as a visibility filter

Graphic3d_ViewOcclusionMask : Provides a way to access occlusion test results for every graphical presentation per each defined view of the viewer

OpenGl_OcclusionQuery: Provides an OpenGL implementation for occlusion queries

Graphic3d_RenderingParams:  provides way to enable or disable occlusion query

AIS_InteractiveContext : resets the view occlusion mask for newly loaded objects

Each view invalidates its occlusion query results if the camera has changed.

vnboccluded: Adds a new DRAW command to return the number of occluded objects in the view

### DIFF
--- a/src/OpenGl/OpenGl_OcclusionQuery.cxx
+++ b/src/OpenGl/OpenGl_OcclusionQuery.cxx
@@ -1,0 +1,110 @@
+// Created on: 2024-02-20
+// Created by: Hossam Ali
+// Copyright (c) 2024 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in
+// OCCT distribution for complete text of the license and disclaimer of any
+// warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <OpenGl_OcclusionQuery.hxx>
+
+#include <OpenGl_Context.hxx>
+#include <OpenGl_GlCore15.hxx>
+
+IMPLEMENT_STANDARD_RTTIEXT(OpenGl_OcclusionQuery, OpenGl_Resource)
+
+// =======================================================================
+// function :
+// purpose  :
+// =======================================================================
+OpenGl_OcclusionQuery::OpenGl_OcclusionQuery(): myID(0), inUse(false), started(false)
+{
+}
+
+// =======================================================================
+// function :
+// purpose  :
+// =======================================================================
+OpenGl_OcclusionQuery::~OpenGl_OcclusionQuery() 
+{
+  Release(NULL);
+}
+// =======================================================================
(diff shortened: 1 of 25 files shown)


Hossam86:
Demo for Draw Test Command

https://github.com/user-attachments/assets/d92fbbb6-4ad2-41bb-b606-5c391095f396

Hossam86:
Another test to clarify potential use cases:

In the viewport, I am visualizing two cones positioned one behind the other.
The left console reports the visible pixel count for each object.
At the beginning of the demo, one cone completely occludes the second one, resulting in 0 pixels reported for the occluded cone. In contrast, a finite number of pixels are reported for the visible cone, indicating that one cone is invisible (0 samples) while the other is visible (finite samples).
As I rotate the camera, the previously occluded cone becomes visible on the screen, and the console reports a finite number of pixels for both cones, indicating that both cones are now visible.
Briefly, this is a low-cost query performed by hardware to determine the number of visible pixels for each object. This information can be used for culling purposes to offload invisible objects from memory or simply for visibility queries 

https://github.com/user-attachments/assets/ae67a327-8d72-49d3-812a-f2c6f0024556

src/OpenGl/OpenGl_OcclusionQuery.cxx
```diff
@@ -0,0 +1,110 @@
...
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <OpenGl_OcclusionQuery.hxx>
+
+#include <OpenGl_Context.hxx>
+#include <OpenGl_GlCore15.hxx>
+
+IMPLEMENT_STANDARD_RTTIEXT(OpenGl_OcclusionQuery, OpenGl_Resource)
+
```
dpasukhi, line 23: Please apply format.path from the GitHub artefact to follow the coding rules.
And additionally remove "// function..." from that header

dpasukhi:
@Hossam86 the occlusion is enabled by default?
Or need special command to enable it?

Hossam86:
@dpasukhi  # Enabling/Disabling Occlusion Query in View Rendering Parameters

By default, the occlusion query is disabled in the initialization of the view. The `OcclusionQueryState` is initialized with the `OcclusionQuery_Off` flag:

`OcclusionQueryState(OcclusionQuery_Off),
`

 To enable the occlusion query, the developer can modify the rendering parameters of the view as follows:
`aView->ChangeRenderingParams().OcclusionQueryState = Graphic3d_RenderingParams::OcclusionQuery_On;
`
or 
`aView->ChangeRenderingParams().OcclusionQueryState = Graphic3d_RenderingParams::OcclusionQuery_NoUpdate;`

dpasukhi:
How it was enabled for the demo?
And the implementation have affect on the performance, what can be a reason?

Hossam86:
@dpasukhi The Draw test command `vnboccluded `can enable it through the same flags in the background for testing.

dpasukhi:
@Hossam86 please create a patch which by default will enable that feature to test it everywhere. (later will be reverted)
Additionally, current version have affect on the global performance degradation, will need to detect where is a performance leak even when the functionality is disabled.

PR #438, opened 2025-03-14
Coding - Migration of Inspector to own repository

dpasukhi:
New repository for Inspector: https://github.com/Open-Cascade-SAS/Inspector
Documentation and all build process are fully migrated.
"tools" folder keep to share information for some time.
Keep building inspector as before to keep control of API changes via GitHub actions.


dpasukhi:
The reason for all team review - do you have any doubts or problem with migration? Or some other ideas about that new migration.
From my point of view, the Inspector is bigger then Sample and not connected to SDK to be part of src folder.
As for tools folder is not clear for the final user.

And separated repo with well-documented product looks much better.
And the testing still are done on GitHub and Jenkins side as a copying of the new repo to compile.

Nathanielnyi:
Good

Nathanielnyi:
Thank you for the gift

PR #495, opened 2025-04-11
Documentation - Fix various typos found in codebase

luzpaz:
Found via codespell


luzpaz:
There are a few fixes that look weird that probably need to be reviewed and refined.

dpasukhi (approved):
Thank you for the fix.
Don't see any issues with changes.

PR #503, opened 2025-04-21
Data Exchange, Gltf Reader - Implement non-uniform scaling in Gltf Im…

ikochetkova:
Apply non-uniform scale factors directly to the triangulation during Gltf import.

The issue came from OCP-1948. Please look there for the model and more information.


dpasukhi:
@ikochetkova I would ask you to create a new bool flag to turn on/ turn off modification of geometry directly.
Additionally, could you please describe the behavior on the flag description, that it will affect only on face level.

All the ongoing remarks will be short and only affect small peace of code, not logic.

dpasukhi:
No new remarks.

PR #518, opened 2025-05-12
Modeling - Bnd_BoundSortBox::Compare fails in some cases

AtheneNoctuaPt:
Class Bnd_BoundSortBox is refactored.
Some cases when Bnd_BoundSortBox could potentially fail to check intersection between boxes are fixed.
Google tests are added for Bnd_BoundSortBox class.


dpasukhi:
C++17 will be marked as default in separated patch. For now keep as is.
I will prepare a C++17 default compiler soon.

dpasukhi (approved):

PR #528, opened 2025-05-13
Foundation Classes - Return value is overridden by OCCT

AtheneNoctuaPt:
- Removed OSD_PerfMeter.h and integrated its functionality directly into OSD_PerfMeter.hxx.
- Replaced legacy performance measurement functions with a modern Stopwatch class utilizing std::chrono for high-resolution timing.
- Updated OSD_PerfMeter to manage stopwatches through a singleton StopwatchStorage class.
- Enhanced meter initialization and management to support shared meters by name.
- Implemented methods for starting, stopping, and printing elapsed time for performance meters.
- Added OSD_PerfMeter_Test.cxx to implement unit tests for OSD_PerfMeter functionality.


src/FoundationClasses/TKernel/OSD/OSD_PerfMeter.hxx
```diff
@@ -16,65 +16,61 @@
...
-      Stop();
-  }
+  //! Outputs the meter data and resets it to initial state.
+  Standard_EXPORT void Kill() const;
 
-private:
-  OSD_PerfMeter(const OSD_PerfMeter&);
-  OSD_PerfMeter& operator=(const OSD_PerfMeter&);
+  //! Prints the data of this meter.
+  Standard_EXPORT std::string Print() const;
```
dpasukhi: std::string is not recommended to be out of header. Only internal usage is accepted.

src/FoundationClasses/TKernel/OSD/OSD_PerfMeter.hxx
```diff
@@ -16,65 +16,61 @@
...
+  //! Prints the data of all meters with non-zero elapsed time.
+  Standard_EXPORT static std::string PrintALL();
 
-protected:
-  int myIMeter;
+  //! Resets all meters.
+  Standard_EXPORT static void ResetALL();
+
+private:
+  std::string myMeterName;
```
dpasukhi: Please use only TCollection_AsciiString

dpasukhi (changes requested):

src/FoundationClasses/TKernel/OSD/OSD_PerfMeter.cxx
```diff
@@ -15,389 +15,366 @@
...
-        creates new meter if there is no such meter
-Returns  :      iMeter if OK, -1 if no such meter and cannot create a new one
-======================================================================*/
-int perf_tick_meter(const char* const MeterName)
+//==================================================================================================
+
+void Stopwatch::Start()
 {
-  int ic = find_meter(MeterName);
+  myStartTime = std::chrono::steady_clock::now();
```
dpasukhi: Migration from OSD_Chronometer to chrono is totally different at this realization.
Need to return to OSD_Chronometer.
The difference is a way of calculation of the time (based on CPU usage of single thread or special timer).

dpasukhi (changes requested):
Need to avoid std::chrono in that task because the difference between thread usage and real clocks.

dpasukhi (approved):

Issue #501, opened 2025-04-18
BRepFilletAPI_MakeFillet: Segfault with two curves and rim


pieterhijma:
### Description

Given a geometry with two curves and a "rim" at the end, performing a fillet with a radius that is too large leads to a segfault.

### Expected Behavior

I would expect an error message.

### Actual Behavior

It leads to a segfault.

<details>
<summary>stacktrace:</summary>
<pre>
#0  0x00007ffff7f4ec24 in opencascade::handle<TopLoc_SListNodeOfItemLocation>::handle (this=0x7fffffff9a18, theHandle=...) at /home/pieter/current/occt-liaison/occt-main/src/FoundationClasses/TKernel/Standard/Standard_Handle.hxx:73
#1  0x00007ffff7f4d1d3 in TopLoc_SListOfItemLocation::TopLoc_SListOfItemLocation (this=0x7fffffff9a18, Other=...) at /home/pieter/current/occt-liaison/occt-main/src/FoundationClasses/TKMath/TopLoc/TopLoc_SListOfItemLocation.hxx:57
#2  0x00007ffff7f4db6d in TopLoc_Location::TopLoc_Location (this=0x7fffffff9a18) at /home/pieter/current/occt-liaison/occt-main/src/FoundationClasses/TKMath/TopLoc/TopLoc_Location.hxx:35
#3  0x00007ffff5eb4fca in TopLoc_Location::Multiplied (this=0x7fffffff9c68, Other=...) at /home/pieter/current/occt-liaison/occt-main/src/FoundationClasses/TKMath/TopLoc/TopLoc_Location.cxx:98
#4  0x00007ffff790dd02 in TopLoc_Location::operator* (this=0x7fffffff9c68, Other=...) at /home/pieter/current/occt-liaison/occt-main/src/FoundationClasses/TKMath/TopLoc/TopLoc_Location.hxx:97
#5  0x00007ffff7553cc6 in BRep_Tool::Surface (F=..., L=...) at /home/pieter/current/occt-liaison/occt-main/src/ModelingData/TKBRep/BRep/BRep_Tool.cxx:79
#6  0x00007ffff756d4c6 in BRepAdaptor_Surface::Initialize (this=0x555555d60020, F=..., Restriction=true) at /home/pieter/current/occt-liaison/occt-main/src/ModelingData/TKBRep/BRepAdaptor/BRepAdaptor_Surface.cxx:75
[...]

PR #532, opened 2025-05-15
Modeling - BRepFilletAPI_MakeFillet Segfault with two curves and rim #501

AtheneNoctuaPt:
- Changed the return type of BRep_Tool::Surface to remove reference for consistency.
- Updated all instances where BRep_Tool::Surface is called to reflect the new return type.
- Added a new test case to reproduce a crash during blend operation with specific geometry.


dpasukhi (changes requested):
BRep_Tool do not validate null shape. It is not initialized face.
For null shape we still can return reference by returning reference to the global handle object.
_
If it is possible by checking on place - please check on place the TopoDS_Shape.IsNull() or not.
Checking of pointer all the time inside getter is not the faster solution, but if no other way - oke. But I assume we can solve it in place of issue.
_
In any cases, please do not modify from reference to copy just because of returning nullptr. Better to use global handle.

dpasukhi (approved):

PR #543, opened 2025-05-17
Coding - Small optimization of StepData_StepReaderData

dpasukhi:
Refactor SetEntityNumbers to use reusable memory pool.
Update SetRecord to avoid extra map searches.


ikochetkova (approved):
No remarks.

PR #535, opened 2025-05-15
Coding, RWGltf_CafWriter - fix gltf dashed edges

elias-bananaz:
## Description

When exporting to GLTF the edges appear dashed:
image

## The Problem
The GLTF writer uses `RWGltf_GltfPrimitiveMode_Lines` as the type for edges. According to [the spec](https://www.khronos.org/opengl/wiki/primitive):
> `GL_LINES`: Vertices 0 and 1 are considered a line. Vertices 2 and 3 are considered a line. And so on.

## The Solution
Use line strips instead:
> `GL_LINE_STRIP`: The adjacent vertices are considered lines. Thus, if you pass n vertices, you will get n-1 lines. 

## Result
image

PR #554, opened 2025-05-20
Data Exchange, GLTF - fix saving edges when Merge Faces is enabled

elias-bananaz:
## Description
When using `SetMergeFaces(true)` with edges, the GLTF exported is invalid. The edges are not saved. Using Draco compression fails as well.

## The Problem
When a document is saved with merged faces, a Compound is created to group together the different faces. But when checking the type of the shape in several places, it's checked for either edge, vertex or defaults to face. That means that a compound of edges is also treated using the FaceIterator.

## The Solution
Look at the underlying shape of a compound. The only possibilities should be: Vertex, Edge, Face, or a Compound that contains exclusively one kind of those.

### Note
This PR undoes the change introduced in https://github.com/Open-Cascade-SAS/OCCT/pull/535. While exchanging `Lines` for `LineStrip` works for unmerged faces, where each edge is added individually, this breaks when using merge faces. Since all edges are now in the same mesh, they are all joined together, resulting in extra edges appearing between the actual edges.

Instead, the approach proposed here is to return to using Lines, but use the indices the fix the gap problem.

### DIFF
--- a/src/DataExchange/TKDEGLTF/RWGltf/RWGltf_CafWriter.cxx
+++ b/src/DataExchange/TKDEGLTF/RWGltf/RWGltf_CafWriter.cxx
@@ -2030,7 +2030,7 @@ void RWGltf_CafWriter::writePrimArray(const RWGltf_GltfFace&         theGltfFace
     switch (theGltfFace.Shape.ShapeType())
     {
       case TopAbs_EDGE:
-        myWriter->Int(RWGltf_GltfPrimitiveMode_LineStrip);
+        myWriter->Int(RWGltf_GltfPrimitiveMode_Lines);
         break;
       case TopAbs_VERTEX:
         myWriter->Int(RWGltf_GltfPrimitiveMode_Points);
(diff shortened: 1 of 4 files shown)


src/DataExchange/TKDEGLTF/RWGltf/RWGltf_CafWriter.cxx
```diff
@@ -287,19 +303,17 @@ Standard_Boolean RWGltf_CafWriter::toSkipShape(const RWMesh_ShapeIterator& theSh
...
-    case TopAbs_COMPOUND:
-    case TopAbs_COMPSOLID:
-    case TopAbs_SOLID:
-    case TopAbs_SHELL:
     case TopAbs_FACE:
       return true;
-    case TopAbs_WIRE:
     case TopAbs_EDGE:
     case TopAbs_VERTEX:
-    default:
```
elias-bananaz, line 301: When checking the face in theGltfFace, we shouldn't expect anything other than Face, Edge, Vertex or Compound, since these are what the specific iterators (Face, Edge, Vertex) are extracting. And Compound is only added in the case of merged faces.

dpasukhi (approved):

PR #561, opened 2025-05-24
Configuration - Fixed issue with CSF variable overwriting

dpasukhi:
Include occt_csf file to define CSF variable before usage.


dpasukhi:
Reported by https://github.com/microsoft/vcpkg/issues/44210

PR #562, opened 2025-05-24
Coding, Data Exchange - Optimize entity graph evaluating

dpasukhi:
Refactor Interface_Graph constructors for improved clarity.
Added memory pools to avoid fragmentation.
Added to reuse containers from iterators directly.


ikochetkova (approved):
No remarks

PR #613, opened 2025-07-13
Foundation Classes - Aligned allocation for Allocator

dpasukhi:
Update NCollection_Allocator to have aligned allocation for big and double types.
Add new method to Standard to reallocate the aligned memory


dpasukhi:
At the moment, no benefits

PR #622, opened 2025-07-18
0033895: fix null surface crash in fixshape

sander-adamson-cloudnc:
ShapeFix_Shape crashes with a segfault when given a shape that contains faces without a surface pointer, for example if the shape is tessellated or triangulated, or contains tessellated / triangulated subshapes.

I added null checks to the `const Handle(Geom_Surface)&` parameter of the ShapeAnalysis_Surface constructor, and to the analogous parameter of the Init method, in each case throwing a Standard_NullObject exception.

Now, ShapeFix_Shape::Perform will throw an exception instead of crashing with a segfault whenever it encounters a face without a surface.

I have CLA 1124.


dpasukhi:
Dear @sander-adamson-cloudnc I see 2 similar PR, which you want to keep?

sander-adamson-cloudnc:
I will make a better PR -- I sent these by accident

sander-adamson-cloudnc:
Is there a particular IR branch I should be targeting the PRs to, or is `master` OK?

PR #624, opened 2025-07-18
0033894: fix null surface crash in unifysamedom

sander-adamson-cloudnc:
ShapeUpgrade_UnifySameDomain crashes with a segfault when given a shape that contains faces without a surface pointer, for example if the shape is tessellated or triangulated, or contains tessellated / triangulated subshapes.

I added null check/fallthrough to ClearRts(const Handle(Geom_Surface)& aSurface) and a null check/exception throw to ShapeUpgrade_UnifySameDomain::IntUnifyFaces, throwing a Standard_NullObject exception.

Now, ShapeUpgrade_UnifySameDomain::Build() will throw an exception instead of crashing with a segfault whenever it encounters a face without a surface.

I have CLA 1124.


dpasukhi:
Thank for the patch.
I will assign for myself and will update a few things in the code.
(Patch will be merged with your name)

Best regards, Dmitrii.

sander-adamson-cloudnc:
Ok thanks Dmitrii!

PR #668, opened 2025-08-12
Modeling - fix crash by ignore invalid TessellatedShell

### LINKED ISSUE #667: Crash when loading step file with invalid entity TESSELLATED_SHELL
### Description

```step
#607638=TESSELLATED_SHELL('',(),$) ;
```

argument 1(start from 0) should be a set with at least one item.

### Expected Behavior

ReadFile() can success by ignore invalid entity?

### Actual Behavior

program crashed when call `cafReader.ReadFile(...)`.

```cpp
STEPCAFControl_Reader cafReader;
cafReader.SetProductMetaMode(true);

spdlog::info("read file, ...");
auto status = cafReader.ReadFile(filename.c_str());
if (status != IFSelect_RetDone) {
    spdlog::error("can not read step file.");
    return 1;
}
spdlog::info("read file, done.");
```

### Operating System

Windows

### Compiler

MSVC

### Bitness

64-bit

### OCCT Version

7.9

### Additional Files

_No response_

wsw0108:
fix #667

### DIFF
--- a/src/DataExchange/TKDESTEP/RWStepVisual/RWStepVisual_RWTessellatedShell.cxx
+++ b/src/DataExchange/TKDESTEP/RWStepVisual/RWStepVisual_RWTessellatedShell.cxx
@@ -56,7 +56,7 @@ void RWStepVisual_RWTessellatedShell::ReadStep(
   {
     Standard_Integer nb0  = theData->NbParams(sub2);
     Standard_Integer num2 = sub2;
-    aItems.Resize(1, nb0, false);
+    aItems->Resize(1, nb0, false);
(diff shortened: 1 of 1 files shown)


dpasukhi:
The recommended patch will be create an empty array all the time.
I will analyze the ReadSubList implementation and probably will go with true return always (for all STEP entities)

dpasukhi:
My apologies, I will fix that issue during current or next week.
Current fix is not optimal and I would fix issue for rest 1000 files :)

Thank you for your contribution. Current issue is more complicated and I would avoid some workaround on place.
If you will have something more to contribute - we always welcome.

PR #683, opened 2025-08-19
Draw, MacOS - remove TKIVtkDraw when VTK_USE_X is OFF

qbisi:
fix build failure on darwin platform when USE_VTK=ON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1017,6 +1017,11 @@ foreach (OCCT_3RDPARTY_LIST ${OCCT_3RDPARTY_CMAKE_LIST})
   OCCT_INCLUDE_CMAKE_FILE (${OCCT_3RDPARTY_LIST})
 endforeach()
 
+if (DEFINED VTK_USE_X AND NOT VTK_USE_X)
+    message (STATUS "Info: TKIVtkDraw toolkits excluded due to VTK has no glx support")
+    list (REMOVE_ITEM BUILD_TOOLKITS TKIVtkDraw)
+endif()
+
 if (USE_TK)
   add_definitions (-DHAVE_TK)
 else()


qbisi:
VTK_USE_X is set in vtk-config.cmake in vtk>=9.4.0, hence the ci failed.

should i replace "NOT VTK_USE_X" with "APPLE" ?

dpasukhi:
Dear @qbisi
Could you explain the reason?
Draw plugins required something unsupported? I probably get the point about gxl. In that case, we probably can proceed with integration.
The question is: before 9.4.0 X was required to be enabled for build?

qbisi:
> Draw plugins required something unsupported?

Generally speaking, we can't compile toolkit TKIVtkDraw when vtk is configured VTK_USE_X=OFF, otherwise, a compilation error
```
/tmp/nix-build-opencascade-occt-7.8.1.drv-0/occt-V7_8_1-bd2a789/src/IVtkDraw/IVtkDraw.cxx:85:12: fatal error: 'vtkXRenderWindowInteract
or.h' file not found
   85 |   #include <vtkXRenderWindowInteractor.h>
```
will raise due to missing of some vtk related header, this header will be provided only when vtk is configured with VTK_USE_X=ON, which is the default value on linux platform (on darwin the default is OFF) for any VTK version.

> The question is: before 9.4.0 X was required to be enabled for build?

However, VTK_USE_X is set in vtk-config.cmake only in vtk>=9.4.0, For vtk<9.4.0, we can not infer from VTK_USE_X to decide if we can build TKIVtkDraw, maybe we should try to test header file "vtkXRenderWindowInteractor.h" in vtk.cmake module.

qbisi:
Hi, is there any way we can valid this pr? I think this pull request is at least doing the right thing for VTK ≥ 9.4, and not causing any issues for VTK < 9.4 According to [repology](https://repology.org/project/vtk/versions), Debian Experimental/Fedora Rawhide/Homebrew/Arch linux/Nixpkgs have already upgraded VTK to the latest 9.5.2.

dpasukhi:
Thank you for reminder. Yes, originally commit looks as a stable. But I need a little deep down to validate it. I will try to review it during next week.
It is workaround and I need to know the price for correct fix.

PR #711, opened 2025-09-08
(Test) Check results without BSpline cache

dpasukhi:


dpasukhi:
TODO: issue with UV bound for Surfaces are not reflect the periodic cases

PR #722, opened 2025-09-14
Remove redundant #include of Poly_CoherentTriangulation.hxx from its own header file

yeicor:


dpasukhi:
General PR: https://github.com/Open-Cascade-SAS/OCCT/pull/739

PR #723, opened 2025-09-16
Visualization - Unexpected moving with AIS_ViewCube

dpasukhi:
Fix AIS_ViewCube rotation instability by replacing Euler angles with quaternions

Replace the problematic Euler angle-based rotation handling in AIS_ViewController
with a mathematically robust quaternion approach to eliminate unexpected PI/2
rotations when transitioning from view cube
orientations to mouse gestures.

Key improvements:
- Add myRotateStartQuaternion member to store camera orientation as quaternion
- Replace unstable Euler angle calculations with direct quaternion operations
- Eliminate gimbal lock singularities that occurred at pitch +-90°
- Maintain numerical stability across all camera orientations
- Remove arbitrary coordinate system transformations and magic number workarounds

The quaternion-based approach ensures smooth,
predictable camera behavior
regardless of the initial orientation set by the view
 cube, resolving the
core mathematical instability that caused erratic
rotations.

### DIFF
--- a/src/Draw/TKViewerTest/ViewerTest/ViewerTest_ViewerCommands.cxx
+++ b/src/Draw/TKViewerTest/ViewerTest/ViewerTest_ViewerCommands.cxx
@@ -6375,6 +6375,106 @@ static Standard_Integer VMoveTo(Draw_Interpretor& theDI,
   return 0;
 }
 
+//=======================================================================
+//function : VMouseButton
+//purpose  : Emulates mouse button event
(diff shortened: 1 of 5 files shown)


dpasukhi:
Working smooth ;)
Ready to review

dpasukhi:
Reverted. Z lock is broken + view rotation is not fully replicate old behavior (it was automatically lock Z and pick ancher point)

PR #752, opened 2025-10-18
Foundation Classes - Optimise AsciiString with pre-defined string

dpasukhi:
* Added modern C++ string handling with string_view support.
  * Expanded string operations and comparison methods with new overloads.
  * Introduced compile-time string literal optimizations for better performance.

### DIFF
--- a/src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
+++ b/src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
@@ -26,6 +26,7 @@
 #include <algorithm>
 #include <cctype>
 #include <cstring>
+#include <functional>
 
 namespace
 {
@@ -278,7 +279,10 @@ void TCollection_AsciiString::AssignCat(const Standard_CString theString,
     return;
 
   // Check if theString points into our own buffer (self-assignment protection)
-  const bool isSelfReference = (theString >= myString && theString < myString + myLength);
+  // Use std::less for well-defined pointer comparison (avoids undefined behavior)
+  const bool isSelfReference =
+    !std::less<const Standard_CString>()(theString, myString)
+    && std::less<const Standard_CString>()(theString, myString + myLength);
 
   if (isSelfReference)
   {
@@ -537,7 +541,10 @@ void TCollection_AsciiString::Insert(const Standard_Integer theWhere,
   const int anInsertIndex = theWhere - 1;
 
   // Check if theString points into our own buffer (self-reference protection)
-  const bool isSelfReference = (theString >= myString && theString < myString + myLength);
+  // Use std::less for well-defined pointer comparison (avoids undefined behavior)
+  const bool isSelfReference =
+    !std::less<const Standard_CString>()(theString, myString)
+    && std::less<const Standard_CString>()(theString, myString + myLength);
 
   if (isSelfReference)
   {
@@ -985,14 +992,12 @@ Standard_Integer TCollection_AsciiString::Search(const Standard_CString theWhat,
   if (theWhatLength == 0)
     return -1;
 
-  int i = 0;
-  while (i <= myLength - theWhatLength)
+  for (int i = 0; i <= myLength - theWhatLength; ++i)
   {
     if (memcmp(myString + i, theWhat, theWhatLength) == 0)
     {
       return i + 1; // Return 1-based position
     }
-    i++;
   }
   return -1;
 }
@@ -1005,14 +1010,12 @@ Standard_Integer TCollection_AsciiString::SearchFromEnd(const Standard_CString t
   if (theWhatLength == 0)
     return -1;
 
-  int i = myLength - theWhatLength;
-  while (i >= 0)
+  for (int i = myLength - theWhatLength; i >= 0; --i)
   {
     if (memcmp(myString + i, theWhat, theWhatLength) == 0)
     {
       return i + 1; // Return 1-based position
     }
-    i--;
   }
   return -1;
 }


src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
```diff
@@ -227,113 +236,143 @@ TCollection_AsciiString::TCollection_AsciiString(const Standard_WideChar* theStr
...
-void TCollection_AsciiString::AssignCat(const Standard_CString theOther)
+void TCollection_AsciiString::AssignCat(const Standard_CString theString,
+                                        const Standard_Integer theLength)
 {
-  if (theOther == NULL)
+  if (theLength == 0 || theString == nullptr)
+    return;
+
+  // Check if theString points into our own buffer (self-assignment protection)
+  const bool isSelfReference = (theString >= myString && theString < myString + myLength);
```
AtheneNoctuaPt: I have some doubts regarding safety of this comparison.
Here we are comparing two pointers. Let's assume that `theString` is pointing to the memory allocated at some random point of execution and completely unrelated to `myString`. According to https://en.cppreference.com/w/cpp/language/operator_comparison.html section "Built-in pointer relational comparison":

> 
Otherwise (both pointers point to objects), the result is defined in terms of a partial order consistent with the following rules: 

Given two different elements high and low of an array such than high has higher subscript than low, if one pointer points to high (or a subobject of high) and the other pointer points to low (or a subobject of low), the former compares greater than the latter.

If one pointer points to an element elem (or to a subobject of elem) of an array, and the other pointer is past the end of the same array, the past-the-end pointer compares greater than the other pointer.

If one pointer points to a complete object, a base class subobject or a member subobject obj (or to a subobject of obj), and the other pointer is past the end of obj, the past-the-end pointer compares greater than the other pointer. 

If the pointers point to different [non-zero-sized](https://en.cppreference.com/w/cpp/language/attributes/no_unique_address.html)(since C++20) non-static data members with the same [member access](https://en.cppreference.com/w/cpp/language/access.html)(until C++23) of the same object of a non-union class type, or to subobjects of such members, recursively, the pointer to the later declared member compares greater than the other pointer.

Otherwise, the result is unspecified.


According to that description this comparison is well defined of `theString` and `myString` are pointing to the same array but **unspecified** if they point to two different char arrays. Given that, we can refer to the "Pointer total order" section:

There exists an implementation-defined strict total order over pointers in each program. The strict total order is consistent with the partial order described above: unspecified results become implementation-defined, while other results stay the same.

Pointer comparison with the strict total order is applied in the following cases:

Calling the operator() of the pointer type specializations of [std::less](https://en.cppreference.com/w/cpp/utility/functional/less.html), [std::greater](https://en.cppreference.com/w/cpp/utility/functional/greater.html), [std::less_equal](https://en.cppreference.com/w/cpp/utility/functional/less_equal.html), and [std::greater_equal](https://en.cppreference.com/w/cpp/utility/functional/greater_equal.html).


So I think, it would only be a safe comparison if use [std::less](https://en.cppreference.com/w/cpp/utility/functional/less.html), [std::greater](https://en.cppreference.com/w/cpp/utility/functional/greater.html), [std::less_equal](https://en.cppreference.com/w/cpp/utility/functional/less_equal.html), and [std::greater_equal](https://en.cppreference.com/w/cpp/utility/functional/greater_equal.html). Assuming, of course, I understood the article correctly.

src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
```diff
@@ -468,246 +501,188 @@ Standard_Integer TCollection_AsciiString::FirstLocationNotInSet(
...
+  if (theLength == 0 || theString == nullptr)
+    return;
+
+  const int anInsertSize  = theLength;
+  const int anOldLength   = myLength;
+  const int aNewLength    = anOldLength + anInsertSize;
+  const int anInsertIndex = theWhere - 1;
+
+  // Check if theString points into our own buffer (self-reference protection)
+  const bool isSelfReference = (theString >= myString && theString < myString + myLength);
```
AtheneNoctuaPt: The same comparison issue mentioned before.

src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
```diff
@@ -972,116 +949,70 @@ void TCollection_AsciiString::Remove(const Standard_Integer where, const Standar
...
+    if (memcmp(myString + i, theWhat, theWhatLength) == 0)
     {
-      k = i++;
-      j = 0;
-      while (j < size && mystring[k++] == swhat[j++])
-        if (j == size)
-          return i;
+      return i + 1; // Return 1-based position
     }
+    i++;
```
AtheneNoctuaPt: No need to do while loop here, can be a for loop:
```
for (int i = 0; i <= myLength - theWhatLength; ++i)
{
  if (memcmp(myString + i, theWhat, theWhatLength) == 0)
  {
    return i + 1; // Return 1-based position
  }
}
```

src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
```diff
@@ -972,116 +949,70 @@ void TCollection_AsciiString::Remove(const Standard_Integer where, const Standar
...
+    return -1;
 
-//=================================================================================================
-
-Standard_Integer TCollection_AsciiString::SearchFromEnd(const TCollection_AsciiString& what) const
-{
-  int size = what.mylength;
-  if (size)
+  int i = myLength - theWhatLength;
+  while (i >= 0)
```
AtheneNoctuaPt: Can also be turned into for loop.

src/FoundationClasses/TKernel/TCollection/TCollection_AsciiString.cxx
```diff
@@ -1286,35 +1205,64 @@ void TCollection_AsciiString::reallocate(const int theLength)
...
+  myString = THE_DEFAULT_CHAR_STRING;
+}
+
+//=================================================================================================
+
+Standard_Boolean TCollection_AsciiString::IsSameString(const Standard_CString theString1,
+                                                       const Standard_Integer theLength1,
+                                                       const Standard_CString theString2,
+                                                       const Standard_Integer theLength2,
+                                                       const Standard_Boolean theIsCaseSensitive)
```
AtheneNoctuaPt, line 1250: Is there any reason this member is non-static?

PR #584, opened 2025-06-24
Shape Healing - Optimize FixFaceOrientation

dpasukhi:
Rework algorithm with efficient data structures and graph-based connectivity analysis

PR #753, opened 2025-10-20
Shape Healing - Regression after #584

dpasukhi:
Fixed issue with unstable shape order after fixing.
Fixed reference data which was changed

### DIFF
--- a/.github/actions/clang-format-check/action.yml
+++ b/.github/actions/clang-format-check/action.yml
@@ -53,10 +53,9 @@ runs:
       run: |
         $files = Get-Content "changed_files.txt" | Where-Object { Test-Path $_ }
         if ($files.Count -gt 0) {
-          # Ensure $files is always an array and properly join for command line
+          # Pass files as individual arguments
           $files = @($files)
-          $fileArgs = ($files | ForEach-Object { "`"$_`"" }) -join ' '
-          python3 .github/actions/scripts/validate-license.py --files $fileArgs --fix
+          & python3 .github/actions/scripts/validate-license.py --files @files --fix
         }
 
     - name: Check formatting


tests/de/step_2/S1
```diff
@@ -1,17 +1,18 @@
...
+puts "TODO CR23096 ALL: CHECKSHAPE : Faulty" 
 
 
 set filename trj12_ttmouse-pe-214.stp
 
 set ref_data {
 DATA        : Faulties = 0  ( 0 )  Warnings = 0  ( 0 )  Summary  = 0  ( 0 )
 TPSTAT      : Faulties = 0  ( 0 )  Warnings = 40  ( 18 )  Summary  = 40  ( 18 )
-CHECKSHAPE  : Wires    = 48  ( 48 )  Faces    = 48  ( 48 )  Shells   = 0  ( 0 )   Solids   = 0 ( 0 )
+CHECKSHAPE  : Wires    = 48  ( 48 )  Faces    = 40  ( 48 )  Shells   = 8  ( 0 )   Solids   = 8 ( 0 )
```
ikochetkova, line 12: Dear [dpasukhi](https://github.com/dpasukhi),

This test case update looks very suspicious in the scope of changes in ShapeFix_Shell. Could you please double-check and add a comment, explaining that this is not a regression?

ikochetkova (commented):
I have no remarks about the code, but please answer a question in test case update before the integration.

tests/de/step_2/S1
dpasukhi: The changes not by current changes, by #584 (I didn't update it on prev master)
It was changes the logic to create shells

PR #765, opened 2025-10-25
Remove CMAKE_BINARY_DIR from CMake files

jordan-bonecutter:
I am including occt as a dep using `FetchContent_Declare`. I had issues with the `CMAKE_BINARY_DIR` being used directly and so I've given it a similar treatment to `OCCT_ROOT_DIR`.


dpasukhi:
Hello.
OCCT is not recommended to use as a submodule. You can take a look into vcpkg integration, which is recommended.
At the moment, that patch will be not merged, because it is only part of necessary changes needed to perform to make positive all workflows to be supported the injection of CMake.
Root dir was required for some internal activities and didn't affect the build process.

My apologies, for now I'm rejecting the PR. Thank you for the patch.
If it is working for you, please continue to use that patch.
But for integration to OCCT it will be postponed.

jordan-bonecutter:
@dpasukhi what is the intended way to build against occt? It was my understanding that the cleanest way to build dependencies in CMake is with submodules

dpasukhi:
> @dpasukhi what is the intended way to build against occt? It was my understanding that the cleanest way to build dependencies in CMake is with submodules

The recommended way - have installed version of OCCT or using package manager, like VCPKG or apt or msys/mingw or brew.
And then connect your application with OCCT as a 3rd-party package (find_package)

jordan-bonecutter:
That’s a shame, building from source is much preferred. Is there anything I can do to get this merged? Otherwise I’ll just point at my fork which should meet my needs

dpasukhi:
At the moment that changes not planned to merged.
Well, building from source can be done once and result of build can be reused (just one step before, you can write bash or sh file to build and prepare an installation)
Vcpkg build from sources right inside CMake also.

PR #779, opened 2025-10-30
Data Exchange, Step Import - Add import of coordinate system connection points for dimensions.

ikochetkova:
Add possibility to retrieve and save the whole coordinate system as a connection point for dimensions. Refactor reading of connection points from STEP.
Refactor Set and Get methods of Dimension XCAF object. Add new exporting data to the test method XDumpDGTs and update the test cases respectively.

### DIFF
--- a/src/DataExchange/TKDESTEP/STEPCAFControl/STEPCAFControl_Reader.cxx
+++ b/src/DataExchange/TKDESTEP/STEPCAFControl/STEPCAFControl_Reader.cxx
@@ -2587,6 +2587,49 @@ void readAnnotation(const Handle(XSControl_TransferReader)& theTR,
 }
 
 //=======================================================================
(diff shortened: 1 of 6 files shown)


src/DataExchange/TKDESTEP/STEPCAFControl/STEPCAFControl_Reader.cxx
```diff
@@ -2586,6 +2586,49 @@ void readAnnotation(const Handle(XSControl_TransferReader)& theTR,
   return;
 }
 
+//=======================================================================
+// function : retrieveConnectionPointFromGISU
+// purpose  : auxiliary method for reading connection points
+//=======================================================================
```
dpasukhi: For internal declaration use doxygen style comments.

PR #820, opened 2025-11-11
Foundation Classes - Migrate Bnd_B2x and Bnd_B3x from macro-template to C++ templates

dpasukhi:


src/FoundationClasses/TKMath/Bnd/Bnd_B3.hxx
```diff
@@ -13,23 +13,298 @@
...
+    return (Abs(aDist.X()) > aHSize.X() || Abs(aDist.Y()) > aHSize.Y()
+            || Abs(aDist.Z()) > aHSize.Z());
+  }
+
+  //! Constant representing a very large value for void box initialization
+  static constexpr RealType THE_RealLast = RealType(1e30);
+
+private:
+  RealType myCenter[3];
+  RealType myHSize[3];
```
VM4Dim, line 154: `myCenter` and `myHSize` used in Salome Mesh and FreeCAD:

https://github.com/SalomePlatform/smesh/blob/V9_15_0/src/StdMeshers/StdMeshers_Adaptive1D.cxx#L76-L78

https://github.com/FreeCAD/FreeCAD/blob/weekly-2025.11.12/src/3rdParty/salomesmesh/src/StdMeshers/StdMeshers_Adaptive1D.cpp#L76-L78

Set `private:` broken it.
I suggest leave `protected:`

dpasukhi: It is not recommended to mark fields as "protected". It leads to unexpected usage, like done in other applications.
I will create public methods for access these methods to read and write.
Unfortunately, migration to 8.0.0 will be with some changes from application side.
_
Path with public methods will be added during a week. Thank you.
Also, I will change the type of fields from C array to std::array

dpasukhi: PR: #838

PR #824, opened 2025-11-13
 Documentation - Fix whitespaces and typos

luzpaz:


luzpaz:
@dpasukhi any chance you can take a look at https://github.com/Open-Cascade-SAS/OCCT/blob/master/src/ModelingAlgorithms/TKHLR/Intrv/Intrv_Interval.hxx and check if it needs formatting ?

dpasukhi:
Mamma Mia!

Who created that has good imagination. Originally, the idea of that design makes sense.
For now better to keep. I probably need to redesign it, but the information is important in that art.

dpasukhi (approved):

PR #833, opened 2025-11-14
Coding - Rework of Math global functions to stl

AtheneNoctuaPt:
Majority of fuctions now simply call same functions from std namespace. Functions that duplicate std namespace functionality are declared depricated.
Calls of depricated functions are replaced with std functions calls.


dpasukhi:
Full testing has no regression.
But for now I would like to ask to migrate the custom code with 
if (magic number) to some helper, to resovle it in the future. Or maybe use old version for now.
@AtheneNoctuaPt

dpasukhi (approved):
Godspeed!

PR #854, opened 2025-11-25
Documentation - Update Assistant guidelines for clarity and consistency

dpasukhi:


.github/copilot-instructions.md
```diff
@@ -19,17 +19,20 @@ This file provides the comprehensive guidance for AI assistants working with the
     - **Correct:**
       ```cpp
       BRepAlgoAPI_Fuse aFuser(theShape1, theShape2);
-      if (aFuser.IsDone()) {
-          auto aResult = aFuser.Shape();
-      } else {
-          // Handle error
+      if (aFuser.IsDone())
+      {
+        auto aResult = aFuser.Shape();
```
AtheneNoctuaPt: I think it is better to use actual type instead of `auto` keyword in hits for AI to avoid polluting code with keyword `auto`.

.github/copilot-instructions.md
```diff
@@ -82,166 +86,356 @@ Open CASCADE Technology (OCCT) is a comprehensive C++ software development platf
...
+// Option 2: NCollection_OccAllocator - with custom memory pool (preferred for performance)
+Handle(NCollection_IncAllocator) anIncAlloc = new NCollection_IncAllocator();
+NCollection_OccAllocator<gp_Pnt> anAllocator(anIncAlloc);
+std::vector<gp_Pnt, NCollection_OccAllocator<gp_Pnt>> aPooledPoints(anAllocator);
+```
+
+### Modern C++ Features
+
+This is a C++17+ codebase. Use modern features:
+- `auto` for type inference
```
AtheneNoctuaPt: IMHO, we should be more restrictive about using `auto` since it may reduce the code readability.
The following cases are encouraged:
- Syntax actually requires auto: templates, structured binding, etc.
- To omit the type of object in range-for loops.
- To omit the type of container iterators. `auto anIter = aCotaner.begin()`
The following cases are discouraged:
- To omit long typenames - Use type aliases if necessary.
- To omit "obvious" return types - What consists "obvious" varies from person to person.
- Simply to avoid the inconvenience of typing the actual type - Readability is more important.
- Pretty much everything else.

Again, the following is just my opinion and is, of course, a subject to discussion.

AtheneNoctuaPt (approved):

PR #855, opened 2025-11-25
Foundation Classes - Enhance B-Spline Curve Computation

dpasukhi:
- Introduced BSplCLib_KnotArrays template for efficient knot and multiplicity management.
- Refactored BSplCLib_Reverse function to utilize std::reverse for improved readability and performance.
- Updated BSplCLib_InsertKnot and BSplCLib_RaiseMultiplicity to use the new knot array structure.
- Added BSplCLib_PolesCoefficients_Bezier and BSplCLib_IncreaseDegree_Bezier functions for better Bezier curve handling.
- Made copy constructor and assignment operator of BSplCLib_EvaluatorFunction deleted to prevent copying.

### DIFF
--- a/src/FoundationClasses/TKMath/BSplCLib/BSplCLib.cxx
+++ b/src/FoundationClasses/TKMath/BSplCLib/BSplCLib.cxx
@@ -344,13 +344,10 @@ Standard_Integer BSplCLib::MaxKnotMult(const Array1OfInteger& Mults,
                                        const Standard_Integer FromK1,
                                        const Standard_Integer ToK2)
 {
-  const int  aLower   = Mults.Lower();
-  const int* pmu      = &Mults(aLower) - aLower;
-  int        aMaxMult = pmu[FromK1];
-
-  for (int i = FromK1 + 1; i <= ToK2; i++)
+  int aMaxMult = Mults(FromK1);
+  for (int anIndex = FromK1 + 1; anIndex <= ToK2; ++anIndex)
   {
-    aMaxMult = std::max(aMaxMult, pmu[i]);
+    aMaxMult = std::max(aMaxMult, Mults(anIndex));
   }
   return aMaxMult;
 }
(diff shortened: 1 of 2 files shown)


src/FoundationClasses/TKMath/BSplCLib/BSplCLib.cxx
```diff
@@ -359,17 +344,15 @@ Standard_Integer BSplCLib::MaxKnotMult(const Array1OfInteger& Mults,
...
-  for (Standard_Integer i = FromK1; i <= ToK2; i++)
+  for (int i = FromK1 + 1; i <= ToK2; i++)
   {
-    if (MaxMult < pmu[i])
-      MaxMult = pmu[i];
+    aMaxMult = std::max(aMaxMult, pmu[i]);
   }
-  return MaxMult;
+  return aMaxMult;
 }
```
AtheneNoctuaPt: So, why are we doing all the pointer arithmetic here in the first place? Just to avoid boundary checking? Seems hacky and unsafe. If this is actually performance-critical we should consider adding methods to containers that allow accessing stored values without boundary check. Otherwise, we can just use normal loop over `Mults` or potentially `std::max_element`.
Same applies to MinKnotMult()
```
// Normal loop
  int aMaxMult = Mults(FromK1);
  for (int anIndex = FromK1 + 1; anIndex <= ToK2; ++anIndex)
  {
    aMaxMult = std::max(aMaxMult, Mults(anIndex));
  }
  return aMaxMult;
  
// Using std::max_element
  const ptrdiff_t aBeginIndex = FromK1 - Mults.Lower();
  const ptrdiff_t aEndIndex   = ToK2 - Mults.Lower() + 1;
  return *std::max_element(std::next(Mults.begin(), aBeginIndex),
                           std::next(Mults.begin(), aEndIndex));
```

src/FoundationClasses/TKMath/BSplCLib/BSplCLib.cxx
```diff
@@ -637,57 +603,55 @@ BSplCLib_MultDistribution BSplCLib::MultForm(const Array1OfInteger& Mults,
...
-    Last  = ToK2;
-  }
-  else
-  {
-    First = ToK2;
-    Last  = FromK1;
-  }
-  if (First + 1 > Mults.Upper())
+  const int aFirst = std::min(FromK1, ToK2);
+  const int aLast  = std::max(FromK1, ToK2);
```
AtheneNoctuaPt: Can potentially be replaced with `auto [First, Last] = std::minmax(FromK1, ToK2);` to avoid extra comparison.

AtheneNoctuaPt (approved):

PR #860, opened 2025-11-28
Modelling - Boolean fuse segfaults on loft

AtheneNoctuaPt:
BOPAlgo_PaveFiller::FillPaves() and CorrectWires() are now skipping edges without geometry.
Minor refactoring of some functions in BOPAlgo_PaveFiller is performed.


dpasukhi:
New test crash SEGFAL on old version with new fixed.
I suppose GTest much better then anything else.

dpasukhi (approved):

Issue #800, opened 2025-11-03
Large memory consumption in BOPAlgo_PaveFiller


lucacarniato:
### Description

I am attempting to perform a boolean cut operation between two `TopoDS_Shape` shapes using the following setup:

- **Left-hand side (`lhs`)**: A single solid shape.
- **Right-hand side (`rhs`)**: A compound consisting of 127 disconnected shapes.

The goal is to subtract the `rhs` from the `lhs`. The result is a new shape composed of **239 sub-shapes**, representing the holes and gaps introduced by the `rhs` into the `lhs`.

The two compounds share several **coplanar faces**, which may influence the behavior of the boolean operation.

### Expected Behavior

Memory consumption should remain within reasonable limits and **not spike excessively** during the operation.

### Actual Behavior

During execution, there is a **significant spike in memory usage** during the call to `pave_filler.Perform()`  
*(see attached image)*.

<img width="504" height="471" alt="Image" src="https://github.com/user-attachments/assets/80110691-560c-43c9-bf9c-273b94905a84" />

### Sample Code or DRAW Tcl Script

#include <BOPAlgo_CellsBuilder.hxx>
#include <BOPAlgo_PaveFiller.hxx>
#include <BRepBndLib.hxx>
#include <BRepTools.hxx>
#include <BRep_Builder.hxx>
#include <Bnd_Box.hxx>

#include <GeomAPI_ProjectPointOnCurve.hxx>
#include <IntTools_Context.hxx>
#include <NCollection_IncAllocator.hxx>
#include <ShapeUpgrade_UnifySameDomain.hxx>

#include <TopExp_Explorer.hxx>

namespace {

std::pair<TopoDS_Shape, TopoDS_Shape> cut(const TopoDS_Shape &lhs, const TopoDS_Shape &rhs)
{
  BOPAlgo_PaveFiller pave_filler;

  TopTools_ListOfShape arguments;
  arguments.Append(lhs);
  arguments.Append(rhs);
  pave_filler.SetArguments(arguments);

  std::cout << "start pave filler perform" << std::endl;
  pave_filler.Perform();
  std::cout << "end pave filler perform" << std::endl;

  if (pave_filler.HasErrors())
  {
    std::ostringstream oss;
    pave_filler.DumpErrors(oss);
  }

  if (pave_filler.HasWarnings())
  {
    std::cerr << std::endl;
    pave_filler.DumpWarnings(std::cerr);
    std::cerr << std::endl;
  }

[...]

PR #861, opened 2025-11-28
Allocate selected variables inside the face-to-face intersection loop

lucacarniato:
Related to Open-Cascade-SAS/OCCT#800

This change moves the allocation of some selected variables, previously at a broader scope, into the face-to-face intersection loop. These variables were cleared on each iteration, and re-creating them within the loop does not affect the algorithm’s behavior. A dedicated incremental allocator is also assigned to these variables and destroyed at each iteration.

For boolean operations on complex geometries, this approach significantly reduces peak memory consumption (by a factor of 3 to 8 depending on the geometry) with only a minor runtime increase of up to 11%.

Summary:
- Move selected variables into the intersection loop
- Use a dedicated incremental allocator for these variables
- Affects memory layout and performance


dpasukhi:
Dear @lucacarniato 

dpasukhi:
My apologies, I redesign the changes and prepere for integration:
https://github.com/Open-Cascade-SAS/OCCT/pull/864

OCCT 7.9.3 is in process and unfortunately, there is no room to wait for CLA and then ask to fix my remarks.

PR #889, opened 2025-12-06
Modeling - Fix thickness operation regression on circle-to-polygon lofts

dpasukhi:
- Fixed CheckMixedContinuity to detect actual mixed concavity (both convex and concave regions) instead of any G1/non-G1 transitions
- Added null edge and PCurve checks in RefEdgeInter to prevent crashes
- Added GTests for BRepOffset_MakeOffset covering various loft scenarios

### DIFF
--- a/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_Analyse.cxx
+++ b/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_Analyse.cxx
@@ -218,17 +218,14 @@ Standard_Boolean CheckMixedContinuity(const TopoDS_Edge&  theEdge,
   Standard_Integer aNbSamples = 23;
 
   // Check for mixed concavity: convex in some regions, concave in others.
-  Standard_Real    aPar;
-  Standard_Real    aDelta = (aLast - aFirst) / (aNbSamples - 1);
-  Standard_Integer i;
-  Standard_Integer aNbConvex  = 0;
-  Standard_Integer aNbConcave = 0;
-  Standard_Integer aNbValid   = 0;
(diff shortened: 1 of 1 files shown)


dpasukhi:
Performance is terrible. Polygonal intersection with sampling is terrible

src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_Analyse.cxx
```diff
@@ -239,49 +241,36 @@ Standard_Boolean CheckMixedContinuity(const TopoDS_Edge&  theEdge,
...
+      Standard_Real anAngle = aCont.C0Value();
+      if (anAngle > M_PI_2 + theAngTol)
+      {
+        aNbConvex++;
+      }
+      else if (anAngle < M_PI_2 - theAngTol)
+      {
+        aNbConcave++;
+      }
     }
```
AtheneNoctuaPt: It seems like we will only check that `aNbConvex > 0 && aNbConcave > 0` later, not the actual numbers. Wouldn't it make sense to turn those variables to boolean?
```
// Before cycle
bool aHasConvex = false;
bool aHasConcave = false;

   // Inside cycle
    if (!aCont.IsG1() && (!aHasConvex || !aHasConcave)) // Can be skipped if both flags are already set.
    {
      const Standard_Real anAngle = aCont.C0Value();
      aHasConvex = aHasConvex || (anAngle > M_PI_2 + theAngTol);
      aHasConcave = aHasConcave || (anAngle < M_PI_2 - theAngTol);
    }
    
// After cycle
aMixedCont = aHasConvex && aHasConcave;
```

pieterhijma:
Thanks for the fix! It works in FreeCAD as well.

PR #892, opened 2025-12-07
Modeling - Enhance periodic curve handling in ChFi3d_Builder

dpasukhi:
- Updated the intersection logic to check for non-null C2dint1 when determining periodicity of 3D curves.
- Improved robustness of the PerformIntersectionAtEnd function to prevent potential issues with null parameters.

### DIFF
--- a/src/ModelingAlgorithms/TKFillet/ChFi3d/ChFi3d_Builder_C1.cxx
+++ b/src/ModelingAlgorithms/TKFillet/ChFi3d/ChFi3d_Builder_C1.cxx
@@ -3097,7 +3097,7 @@ void ChFi3d_Builder::PerformIntersectionAtEnd(const Standard_Integer Index)
           DStr.SetNewSurface(Face[nb - 1], Sfacemoins1);
         }
         //// for periodic 3d curves ////
-        if (cad.IsPeriodic())
+        if (cad.IsPeriodic() && !C2dint1.IsNull())
         {
           gp_Pnt2d                      P2d = BRep_Tool::Parameters(Vtx, Face[0]);
           Geom2dAPI_ProjectPointOnCurve Projector(P2d, C2dint1);


pieterhijma:
Thanks for this PR!

Although there is no longer a crash as described in #690, the blend doesn't result in correct geometry, which is visible in DRAW and in FreeCAD:

screenshot_003

Below a transcript in DRAW with some checks where `Face #3` is the highlighted face in the screenshot:

```
$ DRAWEXE 
Draw[1]> Draw[2]> source curved-base-sweep-curve-opposite-dir.tcl
tolerance ang : 0.01
tolerance 3d  : 0.0001
tolerance 2d  : 1.0000000000000001e-05
fleche        : 0.001
tolblend 0.01 0.0001 1.0000000000000001e-05 0.001

Draw[3]> vdisplay result
Draw[4]> vfit
Draw[5]> vselmode 4 on
Draw[6]> vstate
Locally selected sub-shapes within result:
       Face #3

Draw[7]> checkshape result
On Shape faulty_1 :
BRepCheck_UnorientableShape
Shape faulty_2  on shape faulty_1 :
BRepCheck_NoCurveOnSurface
Shape faulty_3  on shape faulty_1 :
BRepCheck_NoCurveOnSurface

Faulty shapes in variables faulty_1 to faulty_3 
```

Possibly the other related issues give some more handles to investigate.

dpasukhi:
Yes, the output is not valid. The issue were related to crash.
And I saw your one more ticket related to the quality. It will resolve later.
As an option I would recommend to report new ticket with invalid result. Originally it is our part to re-create issue with new description, but you already created one with quality. Probably you can extend existed

pieterhijma:
Totally understand. The PR fixes the crash, so the original issue can be closed. Sure, I've already investigated, so it's easy for me to create a new issue for the quality.

PR #920, opened 2025-12-15
Mesh - Fix point-in-polygon check for CCW polygons in BRepMesh_Delaun

dpasukhi:
The isVertexInsidePolygon method uses the winding number algorithm,
  which computes cumulative angles from the test point to polygon vertices.
  For a point inside a polygon, this sum should be ±2π.

  The gp_Vec2d::Angle method uses an inverted sign convention where
  CCW rotation gives negative angles. For CCW polygons, the total angle
  is -2π, but the original check only accepted +2π.

  Changed the condition from:
    std::abs(Angle2PI - aTotalAng) > Precision::Angular()
  to:
    std::abs(std::abs(aTotalAng) - Angle2PI) > Precision::Angular()

  This correctly handles both CCW (-2π) and CW (+2π) polygon orientations.


dpasukhi:
Issue: https://tracker.dev.opencascade.org/view.php?id=32395

PR #921, opened 2025-12-15
Coding - Fix unnecessary loop iteration in BRepLib::BuildCurve3d

dpasukhi:
BRep_Tool::CurveOnSurface expects Index >= 1, but the loop started
  from ii = 0, resulting in a wasted iteration that always returned
  a null pointer.

Fixes #0032371


dpasukhi:
Ticket: https://tracker.dev.opencascade.org/view.php?id=32371

gkv311:
The same range issue could be observed in `ShapeUpgrade_UnifySameDomain`, which could be detected by throwing exception inside `BRep_Tool::CurveOnSurface()`  (and in similar method nearby).

dpasukhi:
Thank you a lot. I will check them.
I will try to resolve them faster.
Tomorrow I'm going to apply coding refactoring with types (int/double/handle...). Later container direct declarations on the week 
It will leads to merge conflicts mess for all old changes :(

PR #922, opened 2025-12-15
Modeling - Update function handling in GeomFill_ConstrainedFilling and GeomFill_CoonsAlgPatch

dpasukhi:
- Changed the way functions are set in GeomFill_ConstrainedFilling to use SetFunc for improved clarity.
- Corrected parameter usage in GeomFill_CoonsAlgPatch::Value to ensure proper variable assignment for curve evaluation.

### DIFF
--- a/src/ModelingAlgorithms/TKGeomAlgo/GeomFill/GeomFill_ConstrainedFilling.cxx
+++ b/src/ModelingAlgorithms/TKGeomAlgo/GeomFill/GeomFill_ConstrainedFilling.cxx
@@ -434,9 +434,9 @@ void GeomFill_ConstrainedFilling::Init(const occ::handle<GeomFill_Boundary>& B1,
       {
         occ::handle<Law_Function> fu1, fu2;
         ptch->Func(fu1, fu2);
-        fu1 = Law::MixBnd(occ::down_cast<Law_Linear>(fu1));
-        fu2 = Law::MixBnd(occ::down_cast<Law_Linear>(fu2));
-        ptch->Func(fu1, fu2);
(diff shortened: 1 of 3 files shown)


dpasukhi:
Ticket: https://tracker.dev.opencascade.org/view.php?id=29446

AtheneNoctuaPt (approved):

src/ModelingAlgorithms/TKGeomAlgo/GeomFill/GeomFill_ConstrainedFilling.cxx
```diff
@@ -451,9 +451,9 @@ void GeomFill_ConstrainedFilling::Init(const Handle(GeomFill_Boundary)& B1,
       {
         Handle(Law_Function) fu1, fu2;
         ptch->Func(fu1, fu2);
-        fu1 = Law::MixBnd(Handle(Law_Linear)::DownCast(fu1));
-        fu2 = Law::MixBnd(Handle(Law_Linear)::DownCast(fu2));
-        ptch->Func(fu1, fu2);
```
AtheneNoctuaPt: So the description here says "Changed the way functions are set in GeomFill_ConstrainedFilling to use SetFunc for improved clarity." as if changes are only improve clarity. However the actual behavior is changed:
In original version code on line 456 just overwrites new functions in fu1, fu2.
In new code we instead set those new function to ptch.
So the result is: original - functions in ptch remain the same, new - functions in ptch are replaced with new ones. The old behavior is likely erroneous and new code fixes it.
If my understanding is correct, please update the commit description to make sure it is clear that these changes indeed fix the bug and only improve clarity.

Also - make sure to update the code style here. Like occ::handle instead of Handle macro. You know.

AtheneNoctuaPt (changes requested):

PR #928, opened 2025-12-16
Modeling Algorithms - Fix partial torus creation with inverted V range

dpasukhi:
Fixed BRepPrim_OneAxis to correctly handle cases where MeridianValue(VMax).Y()
is less than MeridianValue(VMin).Y() (inverted height ordering). This occurred
with torus V parameter ranges like (90deg, 270deg) where the geometric "top"
corresponds to VMin rather than VMax.

Changes:
- Added helper methods to BRepPrim_OneAxis:
  - IsHeightInverted(): detects when VMax is geometrically below VMin
  - AreHeightsEqual(): detects degenerate case with equal endpoint heights
  - VTopGeometric()/VBottomGeometric(): return V params for geometric top/bottom
- Fixed AxisEdge vertex ordering for inverted heights by swapping 'first' flags
- Fixed StartFace/EndFace orientation for inverted heights
- Fixed StartWire/EndWire edge direction flags for inverted heights
- Added vertex sharing for equal heights case (AxisTopVertex = AxisBottomVertex)
- Modernized code: replaced #define macros with internal enums in anonymous namespace

Known limitation: Equal heights case (e.g., V range 45deg to 135deg where
sin(45)=sin(135)) produces invalid topology due to degenerate AxisEdge.
This is documented in tests with EXPECT_FALSE assertions.

Added comprehensive GTest suite for BRepPrimAPI_MakeTorus covering:
- Full torus creation
- Partial torus with U angle only
- Various V parameter ranges (symmetric, asymmetric, inverted, equal heights)
- Root cause analysis test documenting the height ordering assumption

Fixed 23612/33133

### DIFF
--- a/src/ModelingAlgorithms/TKPrim/BRepPrim/BRepPrim_OneAxis.cxx
+++ b/src/ModelingAlgorithms/TKPrim/BRepPrim/BRepPrim_OneAxis.cxx
@@ -33,6 +33,10 @@
 
 namespace
 {
+//! Multiplier for Precision::Confusion() to separate overlapping pcurves
+//! when heights at VMin and VMax are equal.
+constexpr double THE_PCURVE_OFFSET_FACTOR = 10.0;
+
 //! Vertex array indices.
 enum VertexIndex
 {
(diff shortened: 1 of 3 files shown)


dpasukhi:
Ticket: https://tracker.dev.opencascade.org/view.php?id=33133

PR #979, opened 2026-01-04
Modeling - Implement global span cache for BSpline and Bezier geometry evaluation

dpasukhi:
Implement lazy-loaded global span caches embedded in Geom_BSplineCurve,
Geom_BSplineSurface, Geom_BezierCurve, and Geom_BezierSurface classes.
The cache stores pre-computed Taylor expansion coefficients (Bezier
representation) for all spans, avoiding repeated cache rebuilds during
D0/D1/D2/D3 evaluations.

Key features:
- Contiguous memory layout for all span coefficients using NCollection_Array1
- O(1) span location using hint-based lookup with binary search fallback
- Per-span atomic validity flags for lock-free read checks
- Thread-safe lazy building with std::shared_mutex (concurrent reads)
- Automatic cache invalidation on geometry modification

New cache classes (private .pxx files, not exported):
- Geom_BSplineCurveCache: Multi-span cache for BSpline curves
- Geom_BSplineSurfaceCache: 2D span grid cache for BSpline surfaces
- Geom_BezierCurveCache: Single-span cache for Bezier curves
- Geom_BezierSurfaceCache: Single-span cache for Bezier surfaces

Simplification of dependent classes:
- GeomAdaptor_Curve: Removed BSplCLib_Cache member, RebuildCache(),
  IsBoundary(), IsCacheValid() - now uses Geom_* cache directly
- GeomAdaptor_Surface: Removed BSplSLib_Cache member, RebuildCache() -
  now uses Geom_* cache directly; added boundary parameter normalization
  for all surface types in D0/D1/D2/D3/DN methods
- GeomGridEval_BSplineCurve/Surface: Simplified to delegate to myGeom->D0/D1/D2
- GeomGridEval_BezierCurve/Surface: Simplified to delegate to myGeom->D0/D1/D2

Performance benefits:
- First evaluation in span: O(N*degree) for cache build + O(degree) eval
- Subsequent evaluations in same span: O(degree) only (Horner scheme)
- Grid evaluation patterns benefit most (tessellation, intersection, etc.)
- Unused spans are never computed (lazy loading)

### DIFF
--- a/src/ModelingAlgorithms/TKGeomAlgo/GeomFill/GeomFill_NSections.cxx
+++ b/src/ModelingAlgorithms/TKGeomAlgo/GeomFill/GeomFill_NSections.cxx
@@ -22,6 +22,7 @@
 
 #include <GeomFill_NSections.hxx>
 
+#include <BSplCLib.hxx>
 #include <GCPnts_AbscissaPoint.hxx>
 #include <Geom_BSplineSurface.hxx>
 #include <Geom_Circle.hxx>


dpasukhi:
10-20% faster, but memory fly away.
The next solution will be update Adapter to cache a few last (3 curve, 9 surface)

PR #984, opened 2026-01-06
Foundation Classes - Inherited Standard_Failure from std::exception

dpasukhi:
First patch in iterative renovation of exceptions.

- Updated exception classes to use std::shared_ptr instead of occ::handle
- Removed redundant inclusion of <Standard_Type.hxx> in various header files across the project.
- Removed Set methods for failure and its define template.
- Removed Raise and Rerise static methods.
- Deprecated getting message with old approach, and moving to what()

### DIFF
--- a/src/FoundationClasses/TKernel/Standard/Standard_Failure.cxx
+++ b/src/FoundationClasses/TKernel/Standard/Standard_Failure.cxx
@@ -78,7 +78,7 @@ void Standard_Failure::captureStackTrace()
   if (aStackLength > 0)
   {
     // Limit stack allocation to 64KB to prevent stack overflow
-    const int aStackBufLen = std::min(std::max(aStackLength * 200, 2048), 65536);
+    const int aStackBufLen = std::clamp(aStackLength * 200, 2048, 65536);
     char*     aStackBuffer = (char*)alloca(aStackBufLen);
     if (aStackBuffer != nullptr)
     {
--- a/src/FoundationClasses/TKernel/Standard/Standard_OutOfMemory.cxx
+++ b/src/FoundationClasses/TKernel/Standard/Standard_OutOfMemory.cxx
@@ -40,12 +40,11 @@ const char* Standard_OutOfMemory::what() const noexcept
 void Standard_OutOfMemory::SetMessageString(const char* theMessage)
 {
   // restrict length of the message by buffer size
-  size_t n = (theMessage ? std::min(std::strlen(theMessage), sizeof(myBuffer) - 1) : 0);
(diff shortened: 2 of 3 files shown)


src/FoundationClasses/TKernel/Standard/Standard_Failure.cxx
```diff
@@ -66,184 +59,112 @@ Standard_Failure::StringRef* Standard_Failure::StringRef::copy_message(
...
-      myStackTrace(nullptr)
+void Standard_Failure::captureStackTrace()
 {
   const int aStackLength = Standard_Failure_DefaultStackTraceLength;
   if (aStackLength > 0)
   {
-    int   aStackBufLen = std::max(aStackLength * 200, 2048);
-    char* aStackBuffer = (char*)alloca(aStackBufLen);
+    // Limit stack allocation to 64KB to prevent stack overflow
+    const int aStackBufLen = std::min(std::max(aStackLength * 200, 2048), 65536);
```
AtheneNoctuaPt: `std::clamp` would probably be better here
`const int aStackBufLen = std::clamp(aStackLength * 200, 2048, 65536);`

src/FoundationClasses/TKernel/Standard/Standard_OutOfMemory.cxx
```diff
@@ -44,50 +40,12 @@ const char* Standard_OutOfMemory::GetMessageString() const
 void Standard_OutOfMemory::SetMessageString(const char* theMessage)
 {
   // restrict length of the message by buffer size
-  size_t n = (theMessage ? std::min(strlen(theMessage), sizeof(myBuffer) - 1) : 0);
+  size_t n = (theMessage ? std::min(std::strlen(theMessage), sizeof(myBuffer) - 1) : 0);
```
AtheneNoctuaPt: Should be const and should be named properly.

src/FoundationClasses/TKernel/Standard/Standard_OutOfMemory.cxx
```diff
@@ -44,50 +40,12 @@ const char* Standard_OutOfMemory::GetMessageString() const
...
 
   // first set line end symbol to be safe in case of concurrent call
   myBuffer[n] = '\0';
```
AtheneNoctuaPt: Header says that class is not thread safe, so maybe this comment is incorrect? In any case it is unclear how setting line ending symbol first would help in case of concurrent calls. If you understand what this comment tries to communicate, please update it.

dpasukhi: My apologies, that all is artefacts of one of my tries (some algo which I removed during working on that patch) indeed need to clean up

PR #995, opened 2026-01-12
Modeling - Tangent case with 2 cylinders

dpasukhi:


dpasukhi:
new PR: https://github.com/Open-Cascade-SAS/OCCT/pull/1228

PR #1011, opened 2026-01-16
Use GNUInstallDirs to set installation directories

Zopolis4:
Allow the user to override directories via typical options if necessary, otherwise use the GNU defaults.


dpasukhi:
Dear @Zopolis4 

dpasukhi:
I didn't review, but all variables can be overrided. You just change the way of overriding. You can override all your changed values directly by providing arguments to cmake or override after first configuration.

Zopolis4:
Using GNUInstallDirs allows the right directories to be picked automatically in some cases, such as `lib64` rather than `lib`.

dpasukhi:
Thank you for the contribution.

Unfortunately, we cannot integrate the proposed changes because the CLA has not been signed.

The idea of supporting GNUInstallDirs may be considered separately as a future configuration feature. However, the current implementation requires additional work for multiarch library directories, absolute installation paths, cached option synchronization, installed CMake package paths, and resource discovery.

 We will track the general requirement separately. Thank you for bringing this use case to our attention.

PR #1015, opened 2026-01-18
Foundation Classes - Performance optimizations and new high-performance collections

dpasukhi:
This commit introduces performance improvements across fundamental OCCT classes
and adds new high-performance collection types optimized for modern CPU architectures.

New Collection Classes:
- NCollection_FlatDataMap: High-performance hash map using open addressing with
  Robin Hood hashing. Provides better cache locality than NCollection_DataMap
  by storing all key-value pairs inline in a contiguous array. Features include
  power-of-2 sizing for fast modulo operations, cached hash codes, exception-safe
  insertion, and no per-element memory allocations.
- NCollection_FlatMap: High-performance hash set with the same optimizations.

Matrix and Vector Optimizations:
- math_Matrix: Cache-friendly i-k-j loop order for matrix multiplication.
  The inner loop now accesses matrix rows sequentially, significantly
  improving cache utilization for large matrices.
- math_VectorBase: Norm() and Norm2() rewritten with 4-way loop unrolling
  enabling better SIMD vectorization. Partial sums are combined pairwise for
  improved numerical stability.

Thread Safety Improvements:
- Standard_Transient: Optimized reference counting with explicit memory ordering.
  IncrementRefCounter uses relaxed ordering (sufficient for pure counting).
  DecrementRefCounter uses release ordering with an acquire fence only when
  the count reaches zero, avoiding unnecessary synchronization overhead on
  every decrement (follows std::shared_ptr pattern).
- Standard_Mutex: Deprecated in favor of std::mutex. Added deprecation warnings
  indicating removal in OCCT 8.0.0.

Bug Fixes:
- OSD_Thread (Windows): Added error handling for DuplicateHandle failure in
  Assign() method, properly resetting handle and thread ID on failure.
- OSD_Thread (POSIX): Fixed nanoseconds overflow in Wait() when the computed
  timeout exceeds 1 second. Added normalization to properly carry excess
  nanoseconds to seconds.

Tests:
- Added comprehensive GTest suites for NCollection_FlatDataMap and
  NCollection_FlatMap covering basic operations, iterators, edge cases,
  collisions, and performance characteristics.

### DIFF
--- a/src/FoundationClasses/TKernel/NCollection/NCollection_FlatMap.hxx
+++ b/src/FoundationClasses/TKernel/NCollection/NCollection_FlatMap.hxx
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 OPEN CASCADE SAS
+// Copyright (c) 2026 OPEN CASCADE SAS
 //
 // This file is part of Open CASCADE Technology software library.
 //
@@ -19,6 +19,7 @@
 #include <NCollection_DefaultHasher.hxx>
 
 #include <new>
+#include <optional>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


src/FoundationClasses/TKernel/NCollection/NCollection_FlatMap.hxx
```diff
@@ -0,0 +1,539 @@
...
+ *
+ * @tparam TheKeyType Type of keys
+ * @tparam Hasher     Hash and equality functor (default: NCollection_DefaultHasher)
+ */
+template <class TheKeyType, class Hasher = NCollection_DefaultHasher<TheKeyType>>
+class NCollection_FlatMap
+{
+public:
+  //! STL-compliant typedef for key type
+  typedef TheKeyType key_type;
```
AtheneNoctuaPt: Why do we use typedef instead of using?

src/FoundationClasses/TKernel/NCollection/NCollection_FlatMap.hxx
```diff
@@ -0,0 +1,539 @@
...
+
+private:
+  //! Slot state markers
+  static constexpr uint8_t SLOT_EMPTY   = 0;
+  static constexpr uint8_t SLOT_DELETED = 1;
+  static constexpr uint8_t SLOT_USED    = 2;
```
AtheneNoctuaPt: It seems like this could be enum class instead of just a bunch of constants.
Also, maybe you could provide a description for these markers? It is kinda unclear what's the difference between empty and deleted states and what it affects.

src/FoundationClasses/TKernel/NCollection/NCollection_FlatMap.hxx
```diff
@@ -0,0 +1,539 @@
...
+        {
+          insertImpl(std::move(aOldSlots[i].myKey));
+          aOldSlots[i].myKey.~TheKeyType();
+        }
+      }
+      Standard::Free(aOldSlots);
+    }
+  }
+
+  bool findSlot(const TheKeyType& theKey, size_t& theIndex) const
```
AtheneNoctuaPt: Wouldn't it be better to use std::optional here instead?

Issue #1041, opened 2026-02-01
Boolean fuse operation fails when rotation is involved


skelliam:
### Description

This is related to this FreeCAD issue:  https://github.com/FreeCAD/FreeCAD/issues/27252

While investigating this issue, I found that it comes from the OCCT kernel, so I set out to write a test case in this repo to prove it and hopefully fix it.

Here is a video of the issue in FreeCAD:

![Image](https://github.com/user-attachments/assets/cd76e808-a834-48d0-8441-7305d3f550e0)

I have written some unittests in this repo (PR will be added) that reproduce the issue here on the master branch.  The issue manifests itself slightly differently but there is still a bug here.

Here are the tests with one occurrence (no rotation), two occurrences (180 degree rotation), three occurrences (120 degree rotation), and so on until seven occurrences. I am creating a shape in OCCT like the FreeCAD user, then rotating it and attempting to use `BRepAlgoAPI_Fuse` to fuse the resulting geometry. Then I assert that the surface area of the result should be greater than the surface area of one shape multiplied by (occurrences - 1), for some cases I added a factor 0.80.  When all of the actual surface areas are known this could be changed to assert the actual surface area value.

### Expected Behavior

Expect all fuse operations complete successfully.

### Actual Behavior

The results:

<img width="480" height="157" alt="Image" src="https://github.com/user-attachments/assets/d1ec6eb3-6aba-4184-8544-8ba425360292" />

The test case has an option to enable a .step file dump of the test which is useful for inspection.  Here is what the shapes look like after the fuse operation:

<img width="1661" height="976" alt="Image" src="https://github.com/user-attachments/assets/11b80094-a5c4-4b0b-8fa9-8079341a2d2c" />

### Sample Code or DRAW Tcl Script

I will make a pull request with at least the test cases shortly. 

### Operating System

Windows

### Compiler

MSVC

### Bitness

64-bit

### OCCT Version

latest

PR #1042, opened 2026-02-01
Boolean fuse operation fails when rotation is involved.

skelliam:
DRAFT

Here are some test cases that demonstrate issue #1041. 

Note: I have signed and uploaded the CLA today, 1-Feb-2026.

image


skelliam:
interesting that there are different failures across different platforms... suggesting maybe some kind of floating point issue?

Windows failures:

- ThreeOccurrences120Degrees
- FiveOccurrences72Degrees
- SixOccurrences60Degrees 

Mac with clang failures:

- FiveOccurrences72Degrees 
- SixOccurrences60Degrees 

Linux with clang failures:

- ThreeOccurrences120Degrees 
- FiveOccurrences72Degrees

dpasukhi:
Yes, the impact is very interesting. I will check it in coming days, can be issue with some signal handler for example (I will check on each of os locally)

skelliam:
> Yes, the impact is very interesting. I will check it in coming days, can be issue with some signal handler for example (I will check on each of os locally)

Great, thank you.  I have no idea where to start.  I spent some hours today with Claude and didn't really get anywhere.

Issue #1043, opened 2026-02-02
Unexpected memory footprint differences


lucacarniato:
### Description

The memory footprint in BOPAlgo_PaveFiller::MakeBlocks (BOPAlgo_PaveFiller_6.cxx) was recently improved by introducing a dedicated incremental allocator for variables that can be cleared inside the face‑to‑face loop (see pull request https://github.com/Open-Cascade-SAS/OCCT/pull/864).

This optimization reduced the memory spike by more than 57% for a specific case (from 38.7 GB down to 16.2 GB). However, creating the incremental allocator and associated variables inside the loop reduces the memory footprint even more, from 16 GB to 2.6 GB, without a noticeable performance regression. 

Memory usage was measured before completing each loop iteration, and the maximum value observed was logged, as illustrated in the snippet below.

### Expected Behavior

The memory footprint of the two approaches is expected to be similar. The incremental allocator should reuse the peak memory previously allocated, which should be similar for both approaches.

### Actual Behavior

The memory footprint of the two approaches is very different. Creating the incremental allocator inside the loop reduces the memory footprint much more.

### Sample Code or DRAW Tcl Script

Original code:

`  
occ::handle<NCollection_IncAllocator> aTmpAllocator = new NCollection_IncAllocator;
  
  // Per-iteration collections (use temporary allocator, reset each iteration)
  NCollection_List<int> aLSE(aTmpAllocator), aLBV(aTmpAllocator);
  NCollection_Map<int>  aMVOnIn(100, aTmpAllocator), aMVCommon(100, aTmpAllocator),
    aMVStick(100, aTmpAllocator), aMVEF(100, aTmpAllocator), aMVBounds(100, aTmpAllocator);
  NCollection_IndexedMap<occ::handle<BOPDS_PaveBlock>> aMPBOnIn(100, aTmpAllocator);
  NCollection_Map<occ::handle<BOPDS_PaveBlock>>        aMPBCommon;
  NCollection_DataMap<int, double>                     aMVTol(100, aTmpAllocator);
  NCollection_DataMap<int, NCollection_List<int>>      aDMBV(100, aTmpAllocator);
  
  ...
  
  const int aNbFFPrev = aNbFF;
  for (i = 0; i < aNbFF; ++i, aPS.Next())
  {
    ....
    aMVOnIn.Clear();
    aMVCommon.Clear();
    aMPBOnIn.Clear();
    aMPBCommon.Clear();
    aDMBV.Clear();
[...]

PR #1044, opened 2026-02-02
Modeling - Improve memory management in BOPAlgo_PaveFiller::MakeBlocks

dpasukhi:
This change improves allocator usage in Boolean operations to reduce memory accumulation and make allocator choices explicit.

BOPAlgo_PaveFiller_6.cxx:
- Add aDefaultAllocator for collections requiring proper Free() behavior
- aMVTol, aLPB, aDMVLV now use default allocator (have Remove/UnBind ops)
- aMPBCommon now uses aTmpAllocator (was missing allocator)
- Cross-iteration collections now explicitly use aAllocator
- Add comments explaining allocator choice rationale

BOPDS_DS.cxx:
- SubShapesOnIn() now uses const references instead of copying 4 IndexedMaps
- This eliminates unnecessary memory allocations per face-face intersection

### DIFF
--- a/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_PaveFiller_6.cxx
+++ b/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_PaveFiller_6.cxx
@@ -689,6 +689,7 @@ void BOPAlgo_PaveFiller::MakeBlocks(const Message_ProgressRange& theRange)
   NCollection_Map<int>                                 aMVStick(100, aTmpAllocator);
   NCollection_Map<int>                                 aMVEF(100, aTmpAllocator);
   NCollection_Map<int>                                 aMVBounds(100, aTmpAllocator);
+  NCollection_Map<int>                                 aMI(100, aTmpAllocator);
   NCollection_IndexedMap<occ::handle<BOPDS_PaveBlock>> aMPBOnIn(100, aTmpAllocator);
   NCollection_Map<occ::handle<BOPDS_PaveBlock>>        aMPBCommon(100, aTmpAllocator);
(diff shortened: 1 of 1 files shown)


lucacarniato:
I tested this pull request using the test case referenced in the issue: https://github.com/Open-Cascade-SAS/OCCT/issues/1043 (see the image below).
The two distinct memory peaks are still present. However, when the incremental allocator and its associated variables are moved inside the loop, these peaks disappear.  

Figure_800_1044

dpasukhi:
Dear @lucacarniato it is a pleasure to see your activity :) Really appreciate.
I understand your idea to put allocators into loop, not outside. It will improve avoiding and free previous allocated memory. IncAllocator use "Reset(false)"to reuse memory, once allocated, reuse it.
I assume in some specific case the allocator allocate too much and as a result we keep it for all the time. I will investigate your feedback.

To analyze it better, can you share your code which you calling?

lucacarniato:
Thank you for the quick response. After some additional investigation, I found that creating the `aMI` map using `aTmpAllocator` and clearing it at each loop iteration eliminates the peaks (see the attached patch and the figure below). No differences were found after this change for planar shapes.

`aMI` is of type `NCollection_Map<int>`, similar to the other variables created with `aTmpAllocator`. In the original implementation, it is cleared inside the `GetStickVertices` function, which is invoked at the start of each iteration.

[Create-aMI-using-tmpAllocator.patch](https://github.com/user-attachments/files/25126466/Create-aMI-using-tmpAllocator.patch)

clear-ami

lucacarniato:
Dear @dpasukhi, attached in the comment above the patch that reduces the memory consumption for our case. Please let me know if you need additional information. Thanks.

dpasukhi:
Thank you for your investigation! the patch is applied.
Many thanks, Dmitrii.

PR #1045, opened 2026-02-04
Modeling - Lazily instantiate `BRepTopAdaptor_TopolTool`

ton:
I am looking into reducing OCCTs memory usage for certain Boolean operations. For the case attached in this ticket, a large amount of `IntCurvesFace_Intersector` instances are created for which `ClassifyUVPoint` is never called. This could be avoided by not caching `IntCurvesFace_Intersector` itself in `BRepClass3d_SolidExplorer`, but that feels like a more intrusive change. 

By lazily creating the `BRepTopAdaptor_TopolTool` associated with `IntCurvesFace_Intersector`, peak RSS memory usage is significantly lower in the attached code fragment, where some pattern is cut from a box shape. Peak RSS memory goes down from 110MiB to 93MiB without any performance degradation.

The following program demonstrates the improvement, the input files have been attached to this PR.
[inputs.zip](https://github.com/user-attachments/files/25075755/inputs.zip)

```
#include <BOPAlgo_BOP.hxx>
#include <BRep_Builder.hxx>
#include <BRepTools.hxx>

#include <iostream>

long getRssMibs()
{
  std::ifstream statm("/proc/self/statm");

  long size, resident;
  statm >> size >> resident;

  return (resident * sysconf(_SC_PAGESIZE)) / (1024 * 1024);
}

int main(int, char**)
{
  std::cout << "RSS up front: " << getRssMibs() << '\n';
  BRep_Builder builder;

  TopoDS_Shape box;
  if (!BRepTools::Read(box, "box.brep", builder))
  {
    return EXIT_FAILURE;
  }

  TopoDS_Shape pattern;
  if (!BRepTools::Read(pattern, "pattern.brep", builder))
  {
    return EXIT_FAILURE;
  }

  BOPAlgo_BOP cut;

  cut.SetOperation(BOPAlgo_CUT);
  cut.SetToFillHistory(Standard_False);
  cut.SetCheckInverted(Standard_False);

  cut.AddArgument(box);
  cut.AddTool(pattern);

  cut.Perform();

  if (cut.HasErrors())
  {
    std::cerr << "Failed to perform cut operation\n";
    return EXIT_FAILURE;
  }

  TopoDS_Shape result = cut.Shape();

  std::cout << "RSS after cut: " << getRssMibs() << '\n';

  const bool with_triangulation = false;
  const bool with_normals = false;
  BRepTools::Write(result, "result.brep", with_triangulation, with_normals, TopTools_FormatVersion_VERSION_1);

[...]

PR #1046, opened 2026-02-04
Modeling - Do not always cache `BRepClass3d_SolidClassifier`

ton:
The following PR reduces memory usage for a BOPAlgo cut operation, by only caching `BRepClass3d_SolidClassifier` in `BOPAlgo_BuilderSolid` in case of a hole shell. This reduces the peak RSS memory usage for the same example use case as attached to #1045 from 110MiB to 87MiB on my machine.

The drawback is that in case we are dealing with a hole shell, `BRepClass3d_SolidClassifier` is calculated twice, but for the cases I observed, this has no noticeable impact on performance.


ton:
CLA ID: 1134

dpasukhi:
Dear Ton thank you for your patch. At the moment, 793 is frozen and not planned to be updated until critical security or stability patch.
I will start to port your patch into IR or master branch (inside current PR).

Best regards, Dmitrii.

dpasukhi:
Again, my apologies, after review of your changes, for now the changes I will keep on hold.
It is workaround which I would like to keep to resolve issue more deeper.

Thank you for your investigation. In case if I will not find better solution I will migrate your patch into latest version.

best regards, Dmitrii.

dpasukhi:
Dear @ton

Thank you for your contribution. My apologies for the long delay with integration.

| Median of 7 runs | Old cached | New local | Change |
|---|---:|---:|---:|
| RSS after cut | 172.81 MiB | 151.38 MiB | **−21.44 MiB / −12.4%** |
| Peak RSS | 172.84 MiB | 151.50 MiB | **−21.34 MiB / −12.3%** |
| Runtime | 0.52 s | 0.48 s | no regression |

PR #1075, opened 2026-02-13
Configuration - Fix Cmake static linking warnings missing some spaces

Andrej730:
Hi! Just a small thing - noticed that some spaces were missing in this warning, so "on static" appeared as a single word and no spaces between sentences.

image


dpasukhi:
Thank you for your contribution.

PR #1077, opened 2026-02-13
Coding - Remove redundant null checks before deallocation

dpasukhi:
In C++, delete/delete[] on nullptr and free(NULL) are guaranteed no-ops.
This removes redundant null-check guards before these calls across 39 files, 
 reducing code noise without behavioral change.
Also simplifies map value cleanup in BRepClass3d_SolidExplorer by
  using iterator reference instead of redundant hash lookups.


elfring:
Thanks for your source code improvements.

PR #1100, opened 2026-02-20
Modeling, Hyperbola - Fixing over constraining direction errors

Lokestrom:
The creation of hyperbolas was over constrained in some constructors causing valid hyperbolas to be rejected. In a hyperbola the major radius can be less than the minor radius, therefore this PR removes these incorrect error form these constructors, and adds a collinear check where appropriate.

I will submit the CLA as soon as possible.


dpasukhi:
Thank you for your patch. I will start review in coming days.
My review - I will modify your branch, I'm not expecting to request any changes from your side. After my changes, I will ask you to review modifications.

For your information, I'm on process of modernisation of OCCT and TKGeomBase is coming on the next week.
But I will check your patch before my changes ;)

In meantime if you already registered CLA, would be great to share ID here :)

Lokestrom:
CLA ID: 1137

PR #1164, opened 2026-03-19
Visualization - Include shape transformation in sub-owner bounding box calculation

winzlebee:
I was having an issue where when using AIS_ViewCube, my view transform would go to the vertices I had picked (as a part of a `MeshVS_EntityOwner` selecting vertices) in the local space of the mesh rather than in global space.

This small change in logic (so we transform the returned bounding box) seems to fix this behaviour for this use case. Fitting bounds of selected sub-shapes in `TopoDS_Shape` items seem unaffected.

### DIFF
--- a/src/Visualization/TKV3d/SelectMgr/SelectMgr_SelectableObject.cxx
+++ b/src/Visualization/TKV3d/SelectMgr/SelectMgr_SelectableObject.cxx
@@ -538,6 +538,11 @@ Bnd_Box SelectMgr_SelectableObject::BndBoxOfSelected(
     }
   }
 
+  if (!aBnd.IsVoid() && HasTransformation())
+  {
+    aBnd = aBnd.Transformed(Transformation());
+  }
+
   return aBnd;
 }
 
--- a/tests/vselect/bugs/bug1164_1
+++ b/tests/vselect/bugs/bug1164_1
@@ -1,0 +1,23 @@
+puts "============"
+puts "Verify vfit -selected works correctly for a primitive TopoDS_Shape with transformation"
+puts "============"
+puts ""
+
+pload MODELING VISUALIZATION
+box b 100 200 300
+
+vinit View1
+
+vdisplay b -dispMode 1
+vlocation b -location 500 0 0
+vfit
+
+vzoom 0.5
+vselmode b FACE 1
+vselect 250 125
+
+vfit -selected
+
+if { [vreadpixel 250 250 rgb name] == "BLACK" } {
+    puts "Error: face should be in the middle"
+}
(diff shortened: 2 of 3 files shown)


gkv311:
The current behavior of two methods `AIS_InteractiveObject::BoundingBox()` and `AIS_InteractiveObject::BndBoxOfSelected()` looks inconsistent. I propose moving the fix into `SelectMgr_SelectableObject::BndBoxOfSelected()`.

DRAW scrtipt for testing:
```
pload MODELING VISUALIZATION
box b 100 200 300
vinit View1
vdisplay b -dispMode 1
vlocation b -location 100 0 0
vfit
vzoom 0.5
vselmode b FACE 1
vselect 250 125
vfit -selected
if { [vreadpixel 250 250 rgb name] == "BLACK" } { puts "Error: face should be in the middle" }
```

winzlebee:
> The current behavior of two methods `AIS_InteractiveObject::BoundingBox()` and `AIS_InteractiveObject::BndBoxOfSelected()` looks inconsistent. I propose moving the fix into `SelectMgr_SelectableObject::BndBoxOfSelected()`.
> 
> DRAW scrtipt for testing:
> 
> ```
> pload MODELING VISUALIZATION
> box b 100 200 300
> vinit View1
> vdisplay b -dispMode 1
> vlocation b -location 100 0 0
> vfit
> vzoom 0.5
> vselmode b FACE 1
> vselect 250 125
> vfit -selected
> if { [vreadpixel 250 250 rgb name] == "BLACK" } { puts "Error: face should be in the middle" }
> ```

That makes a lot of sense. Thanks for the DRAW test script. I’ll update the PR and test.

dpasukhi:
CLA signed: submission/523

dpasukhi:
My apologies about the migrate your branch to IR, probably you will need to make pull to push the new test changes.
In case if you want, I can add tests from my side, but before that I will wait for your answer or some time :)

winzlebee:
I've updated the PR to include two DRAW tests, one based on @gkv311's test for TopoDS_Shape and another for MeshVS_Mesh. They fail before and run afterwards, so seems correct.

I've also moved the transformation into `SelectMgr_SelectableObject::BndBoxOfSelected` as this makes more sense.

EDIT: Also I wasn't sure about how to name the DRAW tests. I named them using the `bug` prefix based on this PR ID, as I'm not sure we have a bug in Mantis for this.

dpasukhi:
Many thanks for the patch.
As for tests, indeed it is tricky now. But for the new release I will try to rewrite contributions guidelines.
For now it is oke

PR #1166, opened 2026-03-23
Modeling - New BRrep Graph representation

dpasukhi:


src/ModelingAlgorithms/TKTopAlgo/BRepGraphAlgo/BRepGraphAlgo_BndLib.hxx
```diff
@@ -0,0 +1,170 @@
...
+  //! @param[in] theGraph pre-built BRepGraph
+  //! @param[in,out] theOBB oriented bounding box to fill
+  //! @param[in] theIsTriangulationUsed if true, use triangulation
+  //! @param[in] theIsOptimal if true, use optimal (tighter) computation
+  //! @param[in] theIsShapeToleranceUsed if true, extend by shape tolerances
+  Standard_EXPORT static void AddOBB(const BRepGraph& theGraph,
+                                     Bnd_OBB&         theOBB,
+                                     const bool       theIsTriangulationUsed  = true,
+                                     const bool       theIsOptimal            = false,
+                                     const bool       theIsShapeToleranceUsed = true);
```
AtheneNoctuaPt: Bnd_Box& theBox parameter in those functions is always in a middle. Maybe we should make it first parameter instead just so it would not be in between of input parameters.
Also, can you make those functions to just return the box instead? Or, if pass return value as argument is really needed, at least add methods like, for example, "Get" that wrap "Add" methods and return the box.

src/ModelingData/TKBRep/BRepGraph/BRepGraph_Deduplicate.hxx
```diff
@@ -0,0 +1,84 @@
...
+    int  NbNullifiedCurves    = 0;
+    int  NbHistoryRecords     = 0;
+    bool IsEntityMergeApplied = false;
+
+    //! Topology definition merge counters (active when MergeEntitiesWhenSafe = true).
+    int NbMergedVertices = 0;
+    int NbMergedEdges    = 0;
+    int NbMergedWires    = 0;
+    int NbMergedFaces    = 0;
+
```
AtheneNoctuaPt, line 63: Since it doesn't make sense for those numbers to be negative, maybe we should use some unsigned type for them?

src/ModelingData/TKBRep/BRepGraph/BRepGraph_Builder.hxx
```diff
@@ -0,0 +1,88 @@
...
+  //! @param[in] theParallel    if true, face-level construction runs in parallel
+  static Standard_EXPORT void Perform(BRepGraph&          theGraph,
+                                      const TopoDS_Shape& theShape,
+                                      const bool          theParallel);
+
+  //! Build the full graph with explicit post-pass control.
+  static Standard_EXPORT void Perform(BRepGraph&                            theGraph,
+                                      const TopoDS_Shape&                   theShape,
+                                      const bool                            theParallel,
+                                      const BRepGraphInc_Populate::Options& theOptions);
```
AtheneNoctuaPt, line 45: If we are clearing the graph anyway, wouldn't it make sense to return it from a function instead of accepting it as an argument?

src/ModelingData/TKBRep/BRepGraph/BRepGraph_CacheView.hxx
```diff
@@ -0,0 +1,116 @@
...
+
+//! @brief Non-const view for managing transient cache values on nodes.
+//!
+//! This view is the stable public cache API for BRepGraph callers.
+//! External code should use Cache() for all cache access.
+//! Low-level storage operations such as reserve or cross-graph cache transfer
+//! stay internal to graph-maintenance code through the graph's private cache access.
+//!
+//! Cached values are keyed by BRepGraph_CacheKind descriptors (Handle-based)
+//! and stored as Handle(BRepGraph_CacheValue). Each CacheKind carries a
```
AtheneNoctuaPt, line 27: Handle(BRepGraph_CacheValue) -> occ::handle<BRepGraph_CacheValue>

src/ModelingData/TKBRep/BRepGraph/BRepGraph_ChildExplorer.hxx
```diff
@@ -0,0 +1,227 @@
...
+                                          const TopLoc_Location& theStartLoc,
+                                          TopAbs_Orientation     theStartOri,
+                                          TraversalMode theMode = TraversalMode::DirectChildren);
+
+  [[nodiscard]] bool More() const { return myHasMore; }
+
+  Standard_EXPORT void Next();
+
+  [[nodiscard]] BRepGraph_NodeId Current() const { return myCurrent; }
+
```
AtheneNoctuaPt: Since this is a traversal algorithm, maybe we should add std-like interface (with begin(), end(), etc.) in addition to (or maybe even instead of) More(), Next(), Current()? That way this class will be essentially a range and we can use it with range-for loops and stl algorithms.

src/ModelingData/TKBRep/BRepGraph/BRepGraph_ParentExplorer.hxx
```diff
@@ -0,0 +1,206 @@
...
+    TraversalMode                                theMode = TraversalMode::Recursive);
+
+  //! True if another matching parent is available.
+  [[nodiscard]] bool More() const { return myHasMore; }
+
+  //! Advance to the next matching parent.
+  Standard_EXPORT void Next();
+
+  //! Current matching ancestor node.
+  [[nodiscard]] Standard_EXPORT BRepGraph_NodeId Current() const;
```
AtheneNoctuaPt: Same as for ChildExplorer - having stl-like interface would be great.

PR #1180, opened 2026-04-01
Foundation, Modeling - Fix thread-safety data races for concurrent operations

dpasukhi:
Eliminate non-deterministic SIGSEGV when multiple threads simultaneously create and destroy independent OCCT objects (no shared data between threads). Addresses crashes in BRepCheck_Analyzer, HLRBRep_Algo, BRepPrimAPI_Make*, and TopExp::MapShapesAndAncestors observed under high thread contention (~800 concurrent tests in Swift Testing parallel runner).

Poly_Triangulation: Replace raw Bnd_Box* cache with std::atomic<Bnd_Box*>
plus std::mutex for writes. Lock-free reads via memory_order_acquire; mutex-protected SetCachedMinMax/unsetCachedMinMax prevent use-after-free and double-delete when shared triangulations are accessed from multiple threads.

BRepCheck: Replace broken std::unique_ptr<std::mutex> conditional lock pattern with always-present std::mutex. Use std::unique_lock with defer_lock and myIsParallel guard to avoid mutex overhead in single-threaded mode (zero-cost when parallel=false). Remove fragile double-checked locking in SetParallel; retain parallel flag for sub-algorithm dispatch only.

Foundation globals: Apply std::atomic for simple flags/counters (UnitsMethods_CascadeLengthUnit, OSD_Parallel_ToUseOcctThreads, OSD_WasSetSignal, OSD_SignalStackTraceLength, ADR_ACT_SIGIO_HANDLER, Standard_Failure_DefaultStackTraceLength). Protect Resource_Unicode format state with mutex plus atomic fast-path for lock-free reads after initialization. Protect Plugin::Load cache with std::shared_mutex for concurrent read-only lookups. Protect Units and UnitsAPI globals with std::recursive_mutex for thread-safe lazy initialization. Use std::call_once for OSD_Host Windows init.

TKBool: Convert 19 global mutable static variables to thread_local across 8 files (TopOpeBRepTool_2d, TopOpeBRepDS_connex, TopOpeBRepBuild_HBuilder, TopOpeBRepTool_SC, TopOpeBRepBuild_ffsfs, TopOpeBRepDS_FaceInterferenceTool, TopOpeBRep_vprdeg, TopOpeBRepDS_CurveExplorer) to eliminate cross-thread data corruption in boolean operations.

### DIFF
--- a/src/ModelingAlgorithms/TKBool/TopOpeBRep/TopOpeBRep_vprdeg.cxx
+++ b/src/ModelingAlgorithms/TKBool/TopOpeBRep/TopOpeBRep_vprdeg.cxx
@@ -624,21 +624,21 @@ bool TopOpeBRep_FacesFiller::ProcessVPondgE(const TopOpeBRep_VPointInter&       
   bool        isT2d = false;
   TopoDS_Edge dgEd;
   int         makeI = FUN_putInterfonDegenEd(VP,
-                                             myF1,
-                                             myF2,
-                                             myDataforDegenEd,
-                                             myHDS,
-                                             rankdg,
-                                             dgEd,
-                                             Iiondg,
-                                             T1ondg,
-                                             par1ondg,
-                                             T2ondg,
-                                             par2ondg,
-                                             OOEi,
-                                             parOOEi,
-                                             hasOOEi,
-                                             isT2d);
+                                     myF1,
+                                     myF2,
+                                     myDataforDegenEd,
+                                     myHDS,
+                                     rankdg,
+                                     dgEd,
+                                     Iiondg,
+                                     T1ondg,
+                                     par1ondg,
+                                     T2ondg,
+                                     par2ondg,
+                                     OOEi,
+                                     parOOEi,
+                                     hasOOEi,
+                                     isT2d);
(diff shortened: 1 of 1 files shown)


src/FoundationClasses/TKernel/Resource/Resource_Unicode.cxx
```diff
@@ -583,58 +586,61 @@ bool Resource_Unicode::ConvertUnicodeToANSI(const TCollection_ExtendedString& fr
   return true;
 }
 
-static bool AlreadyRead = false;
+static std::atomic<bool>                AlreadyRead{false};
+static std::atomic<Resource_FormatType> TheFormat{Resource_ANSI};
+static std::mutex                       TheFormatMutex;
 
-static Resource_FormatType& Resource_Current_Format()
+static void readFormatFromConfig()
```
AtheneNoctuaPt, line 593: Previous version of this function checked and updated AlreadyRead flag, current doesn't. Is this intentional?
Also, unlike in SetFormat() we do not acquire mutex here. Again, not sure if it is intentional or not.

PR #1199, opened 2026-04-10
Configuration - Fix BUILD_SOVERSION_NUMBERS=0 handling

angelobartolome:
## Summary

Setting `-DBUILD_SOVERSION_NUMBERS=0` had no effect on macOS (and other non-Windows/Android/Emscripten platforms) due to two bugs:

- **`CMakeLists.txt`**: The guard condition `if (NOT BUILD_SOVERSION_NUMBERS)` evaluates to `true` when the value is `0`, because CMake treats `0` as falsy. This silently overrode the user's explicit `0` back to `2` on every cmake run. Fixed by using `NOT DEFINED ... OR STREQUAL ""` to only apply the platform default when the variable is genuinely absent from the cache.

- **`adm/cmake/occt_toolkit.cmake`**: The `VERSION` target property was set unconditionally regardless of `BUILD_SOVERSION_NUMBERS`. On macOS, setting `VERSION` alone is sufficient for CMake to generate versioned filenames (e.g. `libTKBin.7.9.2.dylib`) and symlinks, even with `SOVERSION` cleared. Fixed by making `VERSION` conditional on `BUILD_SOVERSION_NUMBERS > 0`.

## Test plan

- [x] Configure with `-DBUILD_SOVERSION_NUMBERS=0` and verify only unversioned `.dylib` files are produced (no `libTK*.7*.dylib` variants)
- [x] Configure with `-DBUILD_SOVERSION_NUMBERS=2` (default on macOS/Linux) and verify versioned symlinks are still generated as before
- [ ] Verify Windows/Android/Emscripten default of `0` is still applied when the variable is not explicitly set


dpasukhi:
Dear @angelobartolome 

angelobartolome:
@dpasukhi do you accept digitally signed CLA? (PDF) or does it needs to be actually printed and hand-signed?

dpasukhi:
Yes, it is possible to be digital (most people do that).

dpasukhi (approved):

dpasukhi:
Thank you for your contribution!

PR #1203, opened 2026-04-16
Modeling - Fix null context crash in ShapeUpgrade_FaceDivide::Perform()

### LINKED ISSUE #1179: Thread safety research under concurrent OCCT on arm64 macOS (8.0.0-rc4)
## Environment
- OCCT 8.0.0-rc4 (static library build, arm64 macOS)
- macOS 15.4, Apple M-series
- Concurrent access via Swift Testing parallel test runner (~800 tests running concurrently)

## Observed Behavior
Non-deterministic SIGSEGV when multiple threads simultaneously create and destroy independent OCCT objects. The crash occurs approximately 60-70% of the time when ~800 tests run in parallel. Individual tests always pass when run serially.

The crash point varies between runs but commonly occurs during:
- `HLRBRep_Algo` operations (hidden line removal)
- `TopExp::MapShapesAndAncestors` / `TopTools_IndexedDataMapOfShapeListOfShape` access
- `BRepCheck_Analyzer` validation
- Various `BRepPrimAPI_Make*` constructions

## What We Tried
A standalone C++ reproducer using `dispatch_group_async` with 200 concurrent operations (box/sphere/fuse/fillet/ShapeAnalysis) does **not** reproduce the crash. The issue appears specific to the Swift runtime's thread pool model, which may create higher thread contention than GCD.

## Hypothesis
Some OCCT global or thread-local state (possibly in `NCollection` hash infrastructure, HLR algorithm state, or `BRep_Tool` caches) is not fully thread-safe when independent OCCT operations run on different threads simultaneously. Each thread operates on completely independent shapes with no shared data — the objects are not shared across threads.

## Impact
This affects any application that wraps OCCT in a language with automatic parallel test execution (Swift Testing, async/await workloads). The workaround is serializing all OCCT operations, which negates the parallelism benefits of the BRepGraph work.

## Request
Could the OCCT team consider auditing global/thread-local state for thread safety in the 8.0.0 final release? The BRepGraph discussion mentions parallel processing as a design goal — ensuring the underlying object lifecycle is thread-safe would be valuable.

We're happy to provide additional reproduction details or test on candidate fixes.

gsdali:
## Issue

Fixes https://github.com/Open-Cascade-SAS/OCCT/issues/1179

## Problem

`ShapeUpgrade_FaceDivide::Perform()` crashes with a null pointer dereference in `SplitCurves()` at `Context()->Apply()` because `myContext` is never initialized.

`ShapeUpgrade_FaceDivide` inherits from `ShapeUpgrade_Tool`, which does not allocate a default `ShapeBuild_ReShape` context. The sibling class `ShapeUpgrade_ShapeDivide` handles this correctly with lazy initialization in its own `Perform()`:

```cpp
// ShapeUpgrade_ShapeDivide::Perform() — line 120-121
if (newContext || myContext.IsNull())
    myContext = new ShapeBuild_ReShape;
```

But `ShapeUpgrade_FaceDivide::Perform()` has no such guard.

## Crash details

- **Crash site:** `ShapeUpgrade_FaceDivide::SplitCurves() + 168` — `KERN_INVALID_ADDRESS at 0x0`
- **Reproduces single-threaded** (100% crash rate)
- **Reproduces on Windows x64** (MSVC 19.44) and **macOS arm64** (Apple Clang 21)
- Reproducer: https://github.com/gsdali/OCCTSwift/blob/main/Tests/occt_parallel_crash_portable.cpp
- Windows CI: https://github.com/gsdali/OCCTSwift/actions/runs/24509869728

## Fix

Add lazy context initialization at the top of `Perform()`, matching the pattern used by `ShapeUpgrade_ShapeDivide`:

```cpp
if (Context().IsNull())
```

## Verification

- Single-threaded: was crashing, now passes (100 iterations)
- Parallel (10 worker threads, 300 tasks/round, 10 rounds = 33,000 operations): was crashing 100%, now passes 5/5 runs
- All 11 isolated operation groups pass, plus combined test


dpasukhi:
Dear @gsdali 

_Please validate the filled fields, all fields are requiered, only with exception if written (optional)._

After CLA I will review your code and update the PR. I will clean up comments, remove the QA test and move that test into GTest(C++), not DRAW.

gsdali:
License agreement submitted.

Many Thanks,

--
Edward Lynch-Bell

http://www.linkedin.com/in/edlynchbell/
+61 490 375 565

gsdali:
CLA has been approved and I'm registered under ID 1139

dpasukhi (approved):

PR #1204, opened 2026-04-17
Visualization - Expose SelectMgr pixel tolerance via IVtk picker

soilSpoon:
## Summary

`IVtkTools_ShapePicker::SetTolerance(float)` has been effectively a **dead setter**: it stored the value in `myTolerance` but nothing read it. The actual pick radius was taken from `IVtkOCC_ViewerSelector::myPixTol`, hardcoded to `2` in its constructor, guarded by a one-shot `myToUpdateTol` flag that was flipped to `false` on the first Pick and never reset, and exposed through no public accessor.

As a result, no code path — public API, Draw command, or subclass — could change the pick radius after the first Pick. This matters most on high-DPI displays where 2 physical pixels are a small fraction of a CSS pixel, making edge and vertex selection practically unusable.

## Root cause

`IVtkOCC_ViewerSelector` was introduced in 2011 modeled on `StdSelect_ViewerSelector3d`. It kept its own `myPixTol`/`myToUpdateTol` fields and a `Pick()` override that consulted them, instead of reading the base `SelectMgr_ViewerSelector::myTolerances`. Meanwhile `StdSelect_ViewerSelector3d` was simplified to a typedef over `SelectMgr_ViewerSelector` and AIS now uses the canonical flow at `SelectMgr_ViewerSelector.cxx:1239`:

```cpp
mySelectingVolumeMgr.SetPixelTolerance(myTolerances.Tolerance());
```

`IVtkOCC_ViewerSelector::Pick` never migrated to this pattern, leaving the IVtk picker with a dead external API.

## Change

Align IVtk with the existing AIS tolerance flow:

* Remove `myPixTol` and `myToUpdateTol` from `IVtkOCC_ViewerSelector`.
* Replace the caching branch in all three `Pick` overloads with `mySelectingVolumeMgr.SetPixelTolerance(myTolerances.Tolerance())`, read on every Pick, matching the base `SelectMgr` path.
* Add `SetPixelTolerance(int)` / `PixelTolerance() const` on both `IVtkOCC_ShapePickerAlgo` and `IVtkTools_ShapePicker`, forwarding to `SelectMgr_ViewerSelector`'s public API. Signatures match `SelectMgr_ViewerSelector`'s exactly (same `const int theTolerance` param name).
* **Remove** the dead `SetTolerance(float)` / `GetTolerance()` pair and the unused `myTolerance` field from `IVtkTools_ShapePicker`. Targeting the 8.0 major release window permits this API break; retaining them as deprecated would only preserve misleading no-ops and produce compile-silent bugs.
* Update `ivtkinit` in `TKIVtkDraw` to use the new `SetPixelTolerance(2)` call, matching the historical hardcoded default.

After this change, `IVtkTools_ShapePicker::SetPixelTolerance(n)` behaves exactly like `AIS_InteractiveContext::SetPixelTolerance(n)`.

## Verification

Built vanilla OCCT master (commit `0ebbbedb23`) on macOS and wrote a minimal C++ reproducer that drives `IVtkTools_ShapePicker::Pick` at increasing pixel offsets from a projected box edge. Results (same test, same renderer, only `libTKIVtk` swapped):

| offset from edge | vanilla master | after fix, default | after fix, **`SetPixelTolerance(20)`** |
|---:|:---:|:---:|:---:|
| +0 px | ✅ 1 | ✅ 1 | ✅ 1 |
| +1 px | ✅ 1 | ✅ 1 | ✅ 1 |
| +2 px | ❌ 0 | ✅ 1 | ✅ 1 |
| +3 px | ❌ 0 | ❌ 0 | **✅ 1** |
| +5 px | ❌ 0 | ❌ 0 | **✅ 1** |
| +10 px | ❌ 0 | ❌ 0 | **✅ 2** |
| +20 px | ✅ 1 | ✅ 1 | ✅ 1 |
| +40 px | ❌ 0 | ❌ 0 | ❌ 0 |

Observations:

1. **Default behavior preserved** — base `myTolerances.Tolerance()` defaults produce essentially the same small pick radius as the old `myPixTol = 2`. No regression for existing users.
2. **`SetPixelTolerance(20)` works** — hits at 3, 5, 10 px that previously missed, and `PixelTolerance()` returns `23` (= 20 custom + edge sensitivity contribution from the base `myTolerances` sum), exactly like AIS.
3. **Boundary correct** — offset 40 still misses (off the box). Tolerance does not extend infinitely.
4. **Deletion confirmed empirically** — relinking the test binary against the rebuilt library, the old reference to `SetTolerance(float)` fails with `dyld: Symbol not found: __ZN21IVtkTools_ShapePicker12SetToleranceEf`, i.e. the deletion reached ABI.

## Scope

* 7 files changed, +41 / −57 (net −16 lines of dead caching + dead setter code removed).
* `IVtkOCC_ShapePickerAlgo.{hxx,cxx}`, `IVtkOCC_ViewerSelector.{hxx,cxx}`, `IVtkTools_ShapePicker.{hxx,cxx}`, `IVtkDraw.cxx`.
* No changes to Draw tests or samples in this PR.

## API break notice

The removal of `IVtkTools_ShapePicker::SetTolerance(float)` / `GetTolerance()` is a **source-level break** for downstream consumers that were calling them. Since these methods never affected picking behaviour, the migration is a one-line replacement:

```diff
- picker->SetTolerance(someFloat);
+ picker->SetPixelTolerance(someInt);
```

For downstream code that was relying on `GetTolerance()` to echo back the last value it set, the new `PixelTolerance()` returns the effective pixel tolerance (possibly adjusted by the base `myTolerances` sum).

## Notes

* I could add a Draw command (e.g. `ivtksetpixtol <n>`) and a `tests/vtk/ivtk/` regression script as a follow-up. It was omitted here because `TKIVtkDraw` does not build on macOS (X11-only interactor path in `IVtkDraw_Interactor.cxx`). Happy to add it if preferred — Linux CI would cover it.
* Branch name does not follow the `CR<issue_id>` convention since there is no Mantis ticket; can rename if required.

### DIFF
--- a/src/Draw/TKIVtkDraw/IVtkDraw/IVtkDraw.cxx
+++ b/src/Draw/TKIVtkDraw/IVtkDraw/IVtkDraw.cxx
@@ -363,7 +363,7 @@ void IVtkDraw::ViewerInit(const IVtkWinParams& theParams)
 
     // Init picker
     GetPicker() = vtkSmartPointer<IVtkTools_ShapePicker>::New();
-    GetPicker()->SetTolerance(0.025f);
+    GetPicker()->SetPixelTolerance(2);
     GetPicker()->SetRenderer(GetRenderer());
 
     GetInteractor()->SetShapePicker(GetPicker());
--- a/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ShapePickerAlgo.cxx
+++ b/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ShapePickerAlgo.cxx
@@ -252,6 +252,20 @@ int IVtkOCC_ShapePickerAlgo::NbPicked()
 
 //=================================================================================================
 
+void IVtkOCC_ShapePickerAlgo::SetPixelTolerance(const int theTolerance)
+{
+  myViewerSelector->SetPixelTolerance(theTolerance);
+}
+
+//=================================================================================================
+
+int IVtkOCC_ShapePickerAlgo::PixelTolerance() const
+{
+  return myViewerSelector->PixelTolerance();
+}
+
+//=================================================================================================
+
 bool IVtkOCC_ShapePickerAlgo::processPicked()
 {
   int aNbPicked = myViewerSelector->NbPicked();
--- a/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ShapePickerAlgo.hxx
+++ b/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ShapePickerAlgo.hxx
@@ -40,6 +40,14 @@ public:
   //! Get number of picked entities.
   Standard_EXPORT int NbPicked() override;
 
+  //! Sets the pixel-space selection tolerance used by every subsequent
+  //! Pick call.
+  //! Forwards to SelectMgr_ViewerSelector::SetPixelTolerance.
+  Standard_EXPORT void SetPixelTolerance(const int theTolerance);
+
+  //! Returns the current pixel-space selection tolerance.
+  Standard_EXPORT int PixelTolerance() const;
+
   //! Get activated selection modes for a shape.
   //! @param[in]  theShape a shape with activated selection mode(s)
   //! @return list of active selection modes
--- a/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ViewerSelector.cxx
+++ b/src/Visualization/TKIVtk/IVtkOCC/IVtkOCC_ViewerSelector.cxx
@@ -24,11 +24,7 @@ IMPLEMENT_STANDARD_RTTIEXT(IVtkOCC_ViewerSelector, SelectMgr_ViewerSelector)
 
 //=================================================================================================
 
-IVtkOCC_ViewerSelector::IVtkOCC_ViewerSelector()
-    : myPixTol(2),
-      myToUpdateTol(true)
-{
-}
+IVtkOCC_ViewerSelector::IVtkOCC_ViewerSelector() = default;
 
 //=================================================================================================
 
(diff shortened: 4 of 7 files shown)


dpasukhi:
Dear @soilSpoon 

_Please validate the filled fields, all fields are requiered, only with exception if written (optional)._

soilSpoon:
> Dear @soilSpoon
> 
> Thank you for your patch. To proceed with integration, please complete signing CLA process: The links: [Contribution Guide](https://github.com/Open-Cascade-SAS/OCCT/blob/158ed439c48389bbba0f4f85f10902d9cba96bca/.github/CONTRIBUTING.md) Or https://dev.opencascade.org/get_involved _Please validate the filled fields, all fields are requiered, only with exception if written (optional)._
> 
> In case if you already have signed CLA, please share ID that OCCT team shared with your by email after accepting of CLA.

my CLA id is 1141

soilSpoon:
> > Dear @soilSpoon
> > Thank you for your patch. To proceed with integration, please complete signing CLA process: The links: [Contribution Guide](https://github.com/Open-Cascade-SAS/OCCT/blob/158ed439c48389bbba0f4f85f10902d9cba96bca/.github/CONTRIBUTING.md) Or https://dev.opencascade.org/get_involved _Please validate the filled fields, all fields are requiered, only with exception if written (optional)._
> > In case if you already have signed CLA, please share ID that OCCT team shared with your by email after accepting of CLA.
> 
> my CLA id is 1141



> > Dear @soilSpoon
> > Thank you for your patch. To proceed with integration, please complete signing CLA process: The links: [Contribution Guide](https://github.com/Open-Cascade-SAS/OCCT/blob/158ed439c48389bbba0f4f85f10902d9cba96bca/.github/CONTRIBUTING.md) Or https://dev.opencascade.org/get_involved _Please validate the filled fields, all fields are requiered, only with exception if written (optional)._
> > In case if you already have signed CLA, please share ID that OCCT team shared with your by email after accepting of CLA.

@dpasukhi

> My CLA id is 1141

dpasukhi (approved):

PR #1209, opened 2026-04-19
CI/Build - Enable -Werror on GCC/Clang

jijinbei:
## Motivation
CI previously treated warnings differently across platforms — macOS had `-Werror -Wall -Wextra` injected from the workflow, while Linux and Windows did not. A warning that broke macOS CI often built fine on Linux (and vice versa), making it hard to keep all three green at once. This PR unifies warning handling in the shared CMake flag module so every GCC/Clang build — Linux, macOS, and the MSVC-ClangCL combination — enforces the same set.

## Summary
- Enable `-Werror` on GCC/Clang in a shared CMake block; opt out `-Warray-bounds`, `-Wmaybe-uninitialized`, `-Wstringop-overflow` (GCC 15 false positives in inlined STL).
- Also enable `-Wzero-as-null-pointer-constant` as error.
- Remove now-redundant `-Werror` flags from the macOS CI action.

Fixes surfaced by the stricter build:
- 3 latent uninitialized reads.
- 4 sort/memcpy ranges clamped on fixed-size buffers.
- 8 zero-as-null-pointer sites (macros, X11 `CopyFromParent`/`Aspect_Display` sentinels, bundled `FlexLexer.h`).

Other stricter flags (`-Wold-style-cast`, `-Wnon-virtual-dtor`, etc.) were evaluated but dropped from this PR — the diffs were too large and are better suited to dedicated follow-ups.

## Test plan
- [x] Linux GCC / Linux Clang / macOS Clang / Windows MSVC — CI green.
- [x] GTest pass.

## CLA ID

1142


adm/cmake/occt_defs_flags.cmake
```diff
@@ -44,6 +44,10 @@ else()
     # when using newer Clang versions (LLVM 18+) that are stricter about this C++ standard violation
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-elaborated-enum-base")
     set (CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}   -Wno-elaborated-enum-base")
+    # Apple system headers (declared deprecated but still used) and Clang-18+ cast
```
dpasukhi: Please revert these modification, it is not a good decision to change it in the default cmake

adm/cmake/occt_defs_flags.cmake
```diff
@@ -146,7 +150,17 @@ if (MSVC)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
   endif()
 elseif (CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR (CMAKE_CXX_COMPILER_ID MATCHES "[Cc][Ll][Aa][Nn][Gg]"))
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+  # GCC 15+ -Warray-bounds, -Wmaybe-uninitialized and -Wstringop-overflow produce
+  # false positives inside inlined STL code (e.g. std::sort on fixed-size arrays
```
dpasukhi: Please revert these modification, it is not a good decision to change it in the default cmake

src/DataExchange/TKDEIGES/IGESData/IGESData_IGESWriter.cxx
```diff
@@ -304,7 +304,7 @@ void IGESData_IGESWriter::DirPart(const occ::handle<IGESData_IGESEntity>& anent)
   if (anent->HasShortLabel())
   {
     occ::handle<TCollection_HAsciiString> slab = anent->ShortLabel();
-    for (i = 0; i < slab->Length(); i++)
+    for (i = 0; i < slab->Length() && i < 8; i++)
```
dpasukhi, line 307: why 8?

dpasukhi (changes requested):

src/DataExchange/TKDEIGES/IGESData/IGESData_IGESWriter.cxx
jijinbei: The primary reason was that this loop triggered an `-Warray-bounds` / `-Wstringop-overflow`-class diagnostic once `-Werror` was enabled — the compiler couldn't prove `slab->Length() <= 8` and therefore flagged a  potential write past the 8-byte `label[]` buffer. The `i < 8` clamp makes that bound statically visible and silences the warning.

The `8` itself is not arbitrary: the IGES directory entry's Short Label is a fixed-width 8-column slot
  (positions 56–63 of the DE record), the adjacent `label[8]` buffer is sized to exactly that, and the init loop  just above (`for (i = 0; i < 8; i++) label[i] = snum[i] = ' '`) already hardcodes the same 8.

As a side effect, the clamp also defends against any future `ShortLabel` longer than 8 characters reaching this code — but that wasn't the trigger; the warning was.

adm/cmake/occt_defs_flags.cmake
jijinbei: Apologies — I was leaning too much on "let the compiler enforce the style consistently for every contributor," and lost sight of the impact on downstream library consumers whose builds would suddenly break on any compiler upgrade. You're right that forcing this at the default-CMake level is the wrong tradeoff.

Done in 80bdb6ee5f — the `-Werror` enforcement and related opt-out flags have been moved out of `adm/cmake/occt_defs_flags.cmake` and into the CI workflow (`.github/actions/configure-occt/action.yml`) via `-D CMAKE_CXX_FLAGS`. Downstream consumers now see the same default build behavior as before, while CI keeps the stricter enforcement.

dpasukhi (approved):

dpasukhi:
CLA ID 1142

jijinbei:
thanks!!

PR #1226, opened 2026-04-22
Modelling -  Enhance 3D cycling detection in IntWalk_PWalking

AtheneNoctuaPt:
Fixed hanging in IntWalk_PWalking::Perform() for near-tangent b-spline surfaces with shared boundary edges.

### DIFF
--- a/src/ModelingAlgorithms/TKGeomAlgo/IntWalk/IntWalk_PWalking.cxx
+++ b/src/ModelingAlgorithms/TKGeomAlgo/IntWalk/IntWalk_PWalking.cxx
@@ -1323,8 +1323,7 @@ void IntWalk_PWalking::Perform(const NCollection_Array1<double>& ParDep,
                     // and later returns near it, the intersection curve is closed in 3D.
                     // This catches cases where the UV-based closure check in TestArret fails
                     // (e.g., near-tangent surfaces with shared boundary edges).
-                    else if (line->NbPoints() > 100
-                             && pf.SquareDistance(pl) < aSQDistMax)
+                    else if (line->NbPoints() > 100 && pf.SquareDistance(pl) < aSQDistMax)
                     {
                       close  = true;
                       Arrive = true;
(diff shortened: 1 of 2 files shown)


src/ModelingAlgorithms/TKGeomAlgo/IntWalk/IntWalk_PWalking.cxx
```diff
@@ -1319,6 +1319,18 @@ void IntWalk_PWalking::Perform(const NCollection_Array1<double>& ParDep,
                         bTestFirstPoint = false;
                       }
                     }
+                    // Check for 3D cycling: if the walker has moved away from the starting point
+                    // and later returns near it, the intersection curve is closed in 3D.
+                    // This catches cases where the UV-based closure check in TestArret fails
+                    // (e.g., near-tangent surfaces with shared boundary edges).
+                    else if (line->NbPoints() > 100 && pf.SquareDistance(pl) < aSQDistMax)
```
dpasukhi: why 100? it is some magic number? can we avoid it?

AtheneNoctuaPt: The purpose of 100 was to prevent false closure detection due to numerical oscillation near the start. Basically to check if we made a meaningful amount of steps before starting to check for closure.
I checked this code again and there seems to be variable bTestFirstPoint. We will only execute my new check if it is set to false. And it is set to false precisely when pf.SquareDistance(pl) >= aSQDistMax. Which means the distance between pf and pl was greater then aSQDistMax on at least one of previous iterations. Which seems like better criterion to start checking for closure (and also already implicitly checked). So check for 100 is redundant, I will remove it.
This code is a mess...

dpasukhi:
failed test is not a regression, it is just small difference in 3rd value after dot, need to change TODO in the header of the test to new reference.

dpasukhi (approved):

PR #1232, opened 2026-04-24
Foundation Classes, Math - SIMD-accelerated BVH4/BVH8/BVH16 ray-AABB traversal

jijinbei:
## Summary

Adds runtime-dispatched SIMD ray-vs-AABB kernels and three wide-BVH traversal drivers (BVH4 / BVH8 / BVH16) to the BVH module.

- **19 files added, 0 modified** outside two surgical edits: `BVH_BinaryTree.hxx` gets one new template method (`CollapseToWide<W>`) and a helper, and two `FILES.cmake` listings register the new files.
- **+3601 / −0 lines.**
- **No public API change**, no existing test regresses.

### Dispatch layer

- `BVH_SIMDDispatch.{hxx,cxx}` — runtime CPU detection (cpuid + xgetbv on MSVC, `__builtin_cpu_supports` on GCC/Clang for x86; scalar fallback on ARM and other targets). `BVH::SIMD::Detect()` is computed once per process. `OCCT_BVH_SIMD_FORCE` env var allows forcing a level for testing.
- `BVH_ToolsSIMD_{SSE2,AVX2,AVX512}.hxx` — per-ISA kernel headers, gated with per-function `target` attribute on GCC/Clang so the rest of TKMath stays buildable on hosts without those ISAs.

### BVH4 (uses existing `BVH_QuadTree`)

- `RayBox4_{Scalar,SSE2,AVX2,AVX512}` — 1-ray-vs-4-AABB slab-test kernels.
- `BVH_TraverseQuad.hxx` — `TraverseQuad<>` driver. Inner nodes load up to 4 child AABBs into a SoA record and dispatch to `BVH::SIMD::GetRayBox4()`.

### `BVH_WideTree<int W>` abstraction (BVH8 / BVH16)

- `BVH_WideTree.hxx` — single template tag/specialization replacing what would otherwise be two near-identical ad-hoc classes. `BVH_Tree<T,N,BVH_WideTree<W>>` for `W=8` and `W=16`.
- `BVH_TraverseWide.hxx` — single `TraverseWide<W>` driver covering both fan-outs.
- `BVH_BinaryTree::CollapseToWide<W>()` — single template builder using a new `BVH::GatherDescendants(node, log2(W), …)` helper. `W=8` walks 3 binary levels (2³ = 8 children); `W=16` walks 4 (2⁴ = 16 children).

## SIMD strategy: one natural-fit kernel per width

| BVH width | Natural ISA | Register | Why |
|---|---|---|---|
| BVH4 | SSE2 | xmm 128-bit (4 fp32) | Empirically faster than AVX2 here — the upper 128 bits of ymm sit idle and the 256-bit pipe wakeup cost outweighs the wider scheduling. |
| BVH8 | AVX2 | ymm 256-bit (8 fp32) | First width that saturates ymm. AVX-512 with `__m256` + mask gains only "a few cycles" (measured; not shipped). |
| BVH16 | AVX-512 | zmm 512-bit (16 fp32) | First width that uses the full zmm. Per-lane work genuinely halves vs BVH8/AVX2. |

Mismatched kernels (e.g. SSE2-on-BVH8 with two 4-wide passes, or AVX-512-on-BVH8 keeping zmm half-empty) were intentionally **not** added.

## Tests

385 BVH gtests under `src/FoundationClasses/TKMath/GTests/`:

- **Functional** per width — empty tree, single primitive, all-hit ray, all-miss ray, mosaic-hit ray, ray-behind-origin rejection, parallel-ray-inside-slab.
- **Random consistency** — 1000 random ray + W-AABB cases per kernel, comparing SIMD output to the scalar reference (mask exact, t-values within 1e-3). The AVX-512 BVH16 variant is the acceptance gate for the zmm path.
- **DISABLED microbenchmark** (`BVH_ToolsSIMDBenchmark.DISABLED_RayBox*_KernelComparison`) measuring ns/call per kernel × width.

AVX-512-specific tests `GTEST_SKIP` cleanly on hosts without AVX-512.

### Side-branch cross-validation

In addition to the unit tests above, the equivalence between the BVH4, BVH8 and BVH16 traversals was independently verified on a sibling branch using a 3-way visual cross-check: a DISABLED gtest builds a deterministic random scene of 256 AABBs, runs all three traversals against the same ray, asserts via `EXPECT_EQ` that the three hit sets are identical, and emits a self-contained HTML viewer (Three.js, 3D scene, slider for live N, embedded benchmark table) for inspection. The 3-way `EXPECT_EQ` passed on every host tested. The HTML visualizer is kept on a separate developer-aid branch and is intentionally not included in this PR to keep the diff focused on the kernel/template work.
Screenshot From 2026-04-25 02-40-50

## Performance evidence (measured on Google Cloud)

ns per 1-ray-vs-W-AABB kernel call, 4096 cases × 4096 repeats per kernel.

**Benchmark host:** Google Cloud Compute Engine `c3-highcpu-4` (Intel Xeon Platinum 8481C, Sapphire Rapids, 4 vCPU, `asia-northeast1`) — rented specifically for this PR. All 385 BVH gtests pass on this host; on AVX2-only hosts the 4 AVX-512 tests SKIP cleanly.

| BVH width | Scalar | SSE2 | AVX2 | AVX-512 | Best | Speedup |
|-----------|-------:|-----:|-----:|--------:|------|--------:|
| BVH4      | 31.22  | **3.20** | 3.55 | 3.40    | SSE2     | 9.77×   |
| BVH8      | 75.68  |  —   | **5.70** |  —      | AVX2     | 13.29×  |
| BVH16     | 151.53 |  —   |  —   | **7.47** | AVX-512 | 20.28×  |

(**Bold** = natural-fit kernel for that width; `—` = combinations intentionally not shipped.)

Per-lane (ns), comparing the best kernel of each width:

| Width × kernel | Per-lane | Note |
|---|---:|---|
| BVH16 / AVX-512 | **0.467** | winner — confirms the kernel really uses zmm, not the `__m256`+mask shortcut |
| BVH8 / AVX2     | 0.71  | |
| BVH4 / SSE2     | 0.80  | |

## Compatibility

- **No public API change**, no existing method modified. The only existing files touched are `BVH_BinaryTree.hxx` (one new template method `CollapseToWide<W>` and the `BVH::GatherDescendants` helper) and two `FILES.cmake` listings.
- **Production code does not yet call** the new traversal drivers — they are exercised only by the new tests, leaving callsite migration as opt-in for future PRs.
- **Cross-platform safe**: the `Detect()` function is properly guarded so ARM macOS (Apple Silicon) builds cleanly — `__builtin_cpu_*` is `#if`'d out outside x86, and the fallback path returns `Scalar`.

## Implementation notes

- **Slab formula matches scalar exactly.** All SIMD kernels use `(box - origin) * invDir`, NOT the FMA-friendly form. Parallel-ray edge cases (origin coincides with box face → `0 * inf = NaN`) propagate differently between formulations; the random-consistency tests pin every SIMD kernel to the scalar behaviour.
- **Padding lanes are always-miss boxes.** Inner nodes with fewer than W children get padding lanes filled with `max < min` on every axis, so the SIMD slab test rejects them naturally with no special control flow on the kernel side.
- **`GTEST_SKIP` rather than compile-out** for unsupported ISAs, so the test binary is identical across hosts.

## Build / CI

Verified on the contributor's fork CI matrix: Linux (GCC 13 + Clang 18, AVX2-only host), Windows MSVC (x64 / ARM64), macOS Apple Silicon (Clang, scalar fallback). The full AVX-512 BVH16 zmm path was additionally validated on Google Cloud Sapphire Rapids hardware (see *Performance evidence* above).

## CLA ID

1142


dpasukhi:
Thank you for your contribution and investigation.

 Unfortunately, the proposed SIMD implementation cannot be considered safe across OCCT’s supported compilers, architectures, and CPU configurations. Its integration would require substantial redesign and broader correctness validation, which is beyond the current development cycle.

In any case, thank you for your work.

PR #1262, opened 2026-05-04
Modelling - Fix infinite loop in IntWalk_IWalking::ComputeOpenLine()

AtheneNoctuaPt:
Infinite loop occurred in IntWalk_IWalking::ComputeOpenLine() inside 'while (!Arrive)' loop.
Fixed by checking for degenerate case after very large amount of iterations have passed.


dpasukhi:
The magic limitation with iteration number is not possible to merge with master.
Please consider another solution or drop the fix and proceed with other bug.
@AtheneNoctuaPt

PR #1280, opened 2026-05-20
Perf: IntSurf_LineOn2S → std::vector + hot-path hash-op cleanups

tarik113:
## Summary

Three profile-guided performance changes on top of OCCT 8.0.0, validated with `sample(1)` and a reproducible synthetic Boolean Fuse + meshing benchmark on macOS arm64.

### 1. `IntSurf_LineOn2S`: `NCollection_Sequence` → `std::vector`  *(main change, measurable)*

The `mySeq` member is accessed by index from many hot loops (`IntPatch_WLine::Point(i)` → `IntSurf_LineOn2S::Value(i)` → `mySeq(i)`). `NCollection_Sequence::operator()(int)` keeps an internal current-node cache but degrades to O(N) under non-sequential / interleaved access. In a Boolean Fuse profile, `NCollection_BaseSequence::Find` was the #1 hot symbol.

Swapped the private container to `std::vector<IntSurf_PntOn2S>`. All operations translated 1:1:
- `Length` → `size`, `Reverse` → `std::reverse`, `(i)` → `[i-1]`
- `Append` → `push_back`, `InsertBefore` → `insert`, `Remove`/`RemovePoint` → `erase`
- `Split(i, SS)` reimplemented inline with `insert`+`erase`
- Box-build loops rewritten as range-for

Class API preserved. The allocator-taking constructor signature is kept; the allocator hint is now ignored (`std::vector` uses the default allocator). If embedded users rely on the OCCT allocator pool for `IntSurf_PntOn2S` storage, this can later switch to `std::pmr::vector` with a small adapter.

**Measured:**
- Workload: `BRepAlgoAPI_Fuse` of two 3×3 grids of cylinders+tori (36 non-planar faces per side), serial, deflection ignored, repeated 5×.
- Baseline (`Sequence`): median **4239.8 ms** (range 4220–4257)
- Patched (`vector`): median **4103.5 ms** (range 4068–4126)
- **−136 ms ≈ −3.2%**, variance tighter; behavior identical (same Fuse result, no errors).
- Confirmed in post-patch profile: `NCollection_BaseSequence::Find` is no longer in the top hot symbols.

### 2. `BOPAlgo_PaveFiller_6`: cache `IsClosedFF` in `PerformFF` 4-nested loop *(refactor, behavior-preserving)*

In `PerformFF` (line ~419), the wire/edge/wire/edge nest calls `IsClosedFF(anEdge1, …)` and `IsClosedFF(anEdge2, …)` on every (E1, E2) pair. Both results depend only on a single edge, so:
- `anIsClosed1` lifted out of the W2/E2 loops (computed once per `anEdge1`)
- `anIsClosed2` memoized via `NCollection_DataMap<int,bool> aIsClosedF2Cache` keyed by F2 edge index

`IsClosedFF` is a pure read of the edge's `BRep_CurveRepresentation` list / `BRep_Tool::IsClosed` — safe to cache. Iteration order is preserved, so the `anIsFound` short-circuit fires on the same pair as before.

**Measured:** Call count drops from **4 160 → 1 170 per Fuse** (−71.9%, deterministic counter). Wall-clock change is below noise because each `IsClosedFF` is cheap; the redundant calls were a tiny fraction of the Fuse's overall cost. Net: zero-cost refactor, real-but-invisible work reduction. Kept because future workloads with expensive `IsClosedFF` (edges with many `BRep_CurveRepresentation` entries) may benefit.

### 3. `BRepMesh_BaseMeshAlgo::collectTriangles` + `collectNodes`: drop double hash lookups *(refactor, behavior-preserving)*

Replaced `IsBound(k) + Bind(k, …) + Find(k)` pattern with the idiomatic `ChangeSeek(k) + Bound(k, …)` on the local `DMapOfIntegerInteger`. In `collectNodes`, `IsBound(i) + Find(i)` collapsed to a single `Seek(i)`.

This cuts 1–2 hash ops per node per triangle. Triangle counts are bit-identical against baseline (28 992 triangles produced for both on the test scene).

**Measured:** wall-clock change in mesh benchmark is below noise — `collectTriangles` is not in the meshing profile top (the real meshing hot spots are `NCollection_IncAllocator::allocateSlow`, `NCollection_BaseList::PRemove/PAppend` inside `BRepMesh_DataStructureOfDelaun::cleanLink`, and the Delaunay core itself). Kept because it's a clean idiomatic improvement that costs nothing.

---

## Reviewer notes

- All three changes were validated by direct A/B (same session, same machine, immediate before/after). The first mesh baseline I measured (109 ms) was a cold-machine artifact; the actual baseline once warm was ~52 ms.
- `IntSurf_LineOn2S` triggered a downstream rebuild of ~30 toolkits (TKBO, TKFillet, TKOffset, etc.) — link-only relinks, no source changes elsewhere needed.
- Worth mentioning for context: on the same workload, **toggling the existing `OCCT_Parallel`/`OSD_Parallel` flag on `BRepMesh_IncrementalMesh` gives 3.97× speedup with 10 cores** (90 ms → 23 ms). That is by far the biggest performance lever and requires no patch — just documentation.
- No formal regression / non-regression test on the OCCT test suite was run (only a synthetic Boolean Fuse + meshing micro-benchmark). The `IntSurf_LineOn2S` change is the one that benefits most from running the full OCCT test suite before merge.

## Test plan

- [ ] Run the OCCT non-regression test suite, especially `tests/boolean/*` and `tests/perf/*`
- [ ] Verify `IntSurf_LineOn2S::Split` behavior on edge cases (Index = 1, Index = NbPoints, Index > NbPoints)
- [ ] Measure the patch on a workload with very long walking lines (heavy intersections of high-degree spline surfaces) where Sequence O(N) was actually the bottleneck — synthetic bench here only used cylinders/tori
- [ ] Consider whether the dropped allocator hint in `IntSurf_LineOn2S` constructor matters for any callers using a custom `IntSurf_Allocator`

### DIFF
--- a/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_PaveFiller_6.cxx
+++ b/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_PaveFiller_6.cxx
@@ -415,13 +415,16 @@ void BOPAlgo_PaveFiller::PerformFF(const Message_ProgressRange& theRange)
           BRep_Tool::Triangulation(aF2, aLocation2);
         const bool anIsPlane2 = IsPlaneFF(aSurface2);
 
-        bool anIsFound = false;
+        bool                           anIsFound = false;
+        NCollection_DataMap<int, bool> aIsClosedF2Cache;
         for (TopoDS_Iterator aItW1(aF1); !anIsFound && aItW1.More(); aItW1.Next())
         {
           for (TopoDS_Iterator aItE1(aItW1.Value()); !anIsFound && aItE1.More(); aItE1.Next())
           {
             const TopoDS_Edge& anEdge1      = TopoDS::Edge(aItE1.Value());
             const int          anEdgeIndex1 = myDS->Index(anEdge1);
+            const bool         anIsClosed1 =
+              IsClosedFF(anEdge1, aSurface1, aTriangulation1, aLocation1, anIsPlane1);
 
             for (TopoDS_Iterator aItW2(aF2); !anIsFound && aItW2.More(); aItW2.Next())
             {
@@ -430,10 +433,13 @@ void BOPAlgo_PaveFiller::PerformFF(const Message_ProgressRange& theRange)
                 const TopoDS_Edge& anEdge2      = TopoDS::Edge(aItE2.Value());
                 const int          anEdgeIndex2 = myDS->Index(anEdge2);
 
-                const bool anIsClosed1 =
-                  IsClosedFF(anEdge1, aSurface1, aTriangulation1, aLocation1, anIsPlane1);
-                const bool anIsClosed2 =
-                  IsClosedFF(anEdge2, aSurface2, aTriangulation2, aLocation2, anIsPlane2);
+                bool anIsClosed2;
+                if (!aIsClosedF2Cache.Find(anEdgeIndex2, anIsClosed2))
+                {
+                  anIsClosed2 =
+                    IsClosedFF(anEdge2, aSurface2, aTriangulation2, aLocation2, anIsPlane2);
+                  aIsClosedF2Cache.Bind(anEdgeIndex2, anIsClosed2);
+                }
                 if (!anIsClosed1 && !anIsClosed2)
                 {
                   continue;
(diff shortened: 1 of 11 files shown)


dpasukhi:
Dear @tarik113 

Please avoid std::vector, use NCollection_LinearVector or NCollection_LocalArray + NCollection_Array1.

LinearVector and LocalArray can be used inside Array1 as a pre-allocated array (Array1 can used as span).
For map and DataMap which stored only integer/bool/char values prefer to use FlatMap and FlatDataMap

PR #1292, opened 2026-06-03
Modeling Algorithms - rework Gordon surface construction

dpasukhi:
Remove GeomFill_GordonBuilder and Approx_BSplineApproxInterp.

These classes had implementation similarities with TiGL occ_gordon, which is distributed under the Apache License 2.0.  The code has been removed to avoid ambiguous third-party provenance in this OCCT contribution.

The new implementation converts and reparametrizes input curves with local
Gordon utilities, computes the profile/guide intersection table, normalizes
curve ordering, and prepares compatible B-spline curve families before surface
construction.

GeomFill_NetworkSurface now builds the compatible network directly with OCCT
B-spline tools. Profile and guide skins are interpolated from curve poles, an
intersection-grid reference surface is built from the resolved network points,
the surfaces are aligned to a common knot basis, and the final poles are
corrected without point-grid surface approximation.

Add explicit result statuses to GeomFill_Gordon and GeomFill_NetworkSurface so
callers can distinguish invalid input, preprocessing failures, compatibility
failures, and final construction failures while preserving the IsDone()/Surface()
usage pattern.

Extend Gordon tests with status checks and closed-network coverage for the
low-level network surface builder.


dpasukhi:
[gordon_preview_brep.zip](https://github.com/user-attachments/files/28556845/gordon_preview_brep.zip)

PR #1336, opened 2026-07-11
Foundation Classes - Fix bucket array leak in Poly_MergeNodesTool

mbait:
## Summary

`Poly_MergeNodesTool::MergedNodesMap` derives from `NCollection_BaseMap` but defines no destructor. The base destructor is `= default` and relies on each subclass calling `Destroy()` to free the hash bucket arrays (this is what `NCollection_DataMap` et al. do via `Clear(true)`), so the bucket array allocated by `NCollection_BaseMap::BeginResize()` leaks on every `Poly_MergeNodesTool` instance — i.e. on every STL/OBJ/mesh read that merges nodes (`RWStl_Reader`, `RWObj`).

The leaked size scales with mesh size: `2 * nbFacets` buckets when the facet count is known, or the 995329-bucket default (~8 MB per import on 64-bit) when it is not. The map *nodes* live in the handle-held `NCollection_IncAllocator` and are released correctly — only the raw bucket array leaks.

## LeakSanitizer evidence

Reported downstream against prebuilt V8_0_0_p1 Release libs (farfield-ru/libocct-build#1); resolved stack (LSan with `fast_unwind_on_malloc=0`):

```
#1 Standard::Allocate(unsigned long)                       libTKernel
#2 NCollection_BaseMap::BeginResize(...)                   libTKernel
#3 Poly_MergeNodesTool::MergedNodesMap::Bind(...)          libTKMath
#4 Poly_MergeNodesTool::PushLastElement(int)               libTKMath
#5 RWStl_Reader::ReadAscii(...)                            libTKDESTL
#6 RWStl_Reader::Read(...)                                 libTKDESTL
#7 RWStl::ReadFile(...)                                    libTKDESTL
#8 StlAPI_Reader::Read(TopoDS_Shape&, char const*)         libTKDESTL
```

A one-triangle ASCII STL leaks 7,962,640 bytes per `StlAPI_Reader::Read()` call (the unknown-size default path).

## Fix

Add the missing destructor calling `Destroy(DataMapNode::delNode, true)`, matching the idiom used by the other `NCollection` map classes.

## Verification

ASan/LSan-instrumented host performing `StlAPI_Reader::Read()` + `BRepBndLib::Add()`:
- before: `Direct leak of 7962640 byte(s) in 1 object(s)` with the stack above, nonzero exit;
- after: no LSan report, exit 0.

Also rebuilt Release/Debug/RelWithDebInfo shared-library configurations on Linux (GCC 13) and Windows (MSVC, Ninja) with this change applied on top of V8_0_0_p1 — no build or functional regressions observed in STEP/IGES/STL/OBJ import paths.


dpasukhi:
Dear @mbait thank you for your investigation and detection of leak!

mbait:
I've submitted the signed CLA, waiting for approval.

dpasukhi:
8.1.0 version target is only means that it will be not included into the maintenance releases because of ABI changes.

Thank you for the patch, it will be merged after 8.0.1 (in the beginning of august)

mbait:
So did you get the approved CLA?

dpasukhi:
I'm pinging manager ;(

He for some unknown to me reason didn't processed last CLAs :(
I'm trying to push him, I ask him today is deadline.

I would expect that in coming hours he will send email to you. I also waiting for that :(

dpasukhi:
My apologies for possible delays

dpasukhi:
Dear @mbait

I pre-check your CLA, could you please update the next field:
`Please place an “X” on one of the applicable statements below. Please do NOT mark both statements:`
(it is under your name).

After that your CLA will be pre-approved by me and manager (I hope soon will give you signed copy and ID)

dpasukhi:
@mbait thank you, I provided to the manager directly (removed your pdf for personal data protection reason)

dpasukhi (approved):

dpasukhi:
Dear @mbait,

Thank you for identifying and fixing this leak. We confirm that the issue and the proposed modification are valid for OCCT 8.0.0.

However, the current master branch already resolves the leak in NCollection_BaseMap, whose destructor now releases the bucket storage directly. Applying this modification on top would duplicate the cleanup and introduce an unnecessary traversal of all map buckets.

We will close this PR as already resolved in master.

In any case, thank you for your contribution.

PR #1373, opened 2026-07-16
Modeling - Optimize elementary deviations in GeomLib_CheckCurveOnSurface

### LINKED ISSUE #1372: Modeling - GeomLib_CheckCurveOnSurface spends most of its time re-discovering flat deviation functions
### Description

`GeomLib_CheckCurveOnSurface` (the same-parameter deviation check used by boolean operations via `IntTools_Tools::ComputeTolerance`, and by `BRepLib_ValidateEdge`, `BRepCheck_Edge`, `BRepFill_Sweep`) evaluates the max deviation between a 3D curve and its curve-on-surface composition with, per knot sub-interval:

1. a `math_PSO` particle swarm (3×N control-point seeds + ~100 swarm iterations, ≈310 curve evaluations). The velocity-based early exit never triggers on a flat deviation function: no particle ever *strictly* improves its best value, so the personal/social attraction terms keep velocities oscillating for the full iteration budget;
2. a `math_NewtonMinimum` polish, which **always fails** on a flat function (the Hessian of a constant is singular);
3. the failure fallback: a second full ~310-evaluation PSO on a range 1000× smaller.

For analytic intersections (line/circle/ellipse on plane/cylinder/cone/sphere/torus — the dominant case in prismatic mechanical CAD), the deviation function is numerically zero everywhere: ~620 evaluations per sub-interval to measure ≈1e-16.

Callgrind on a representative workload (128×128×5 baseplate box cut by 9 rounded-box tools in one `BRepAlgoAPI_Cut`, serial mode): `GeomLib_CheckCurveOnSurface::Perform` is **17.9% of the entire boolean operation** (466M of 2 612M instructions), all under `IntTools_FaceFace::ComputeTolReached3d`; 495 728 `TargetFunc::Value` evaluations for 12 boolean ops; `math_NewtonMinimum` failed in 792 of 792 calls, each triggering the fallback swarm.

### Use Case

Any boolean-heavy modeling on analytic geometry; especially single-threaded consumers (WebAssembly builds), where `OSD_Parallel` cannot hide the cost of this check.

### Benefits

Proposed fix (PR to follow): skip the swarm/Newton pipeline when the control-point sampling already shows the deviation is below the geometric noise floor (`Precision::SquareConfusion()` on squared distances) — immediately for fully analytic compositions, after a half-step-offset confirmation resample otherwise.

Measured (median of 15, single-threaded):

| workload | before | after |
|---|---|---|
| baseplate cut, 9 rounded tools, native gcc13 -O2 | 59.5 ms | 47.4 ms (−20.3%) |
| 5-solid fuse (`BRepAlgoAPI_BuilderAlgo`), native | 28.0 ms | 23.2 ms (−17.1%) |
| same workloads, WebAssembly (Emscripten 5.0.3) | 100.9 / 51.2 ms | 80.8 / 41.4 ms (−19.9% / −19.1%) |

Result shapes are bit-identical (topology counts, volume/area, tolerances, `BRepCheck_Analyzer`); DRAW test groups `boolean`, `blend`, `lowalgos` produce identical outcome sets with and without the change.

### Additional Context

A related follow-up candidate (not part of the fix): `math_PSO` could use a stagnation-based termination (no global-best improvement over K steps) so *all* its users stop early on flat objectives; the current velocity criterion cannot fire in that regime.

andymai:
Fixes #1372.

`GeomLib_CheckCurveOnSurface` runs, per sub-interval: a ~310-evaluation `math_PSO` swarm, a `math_NewtonMinimum` polish that is guaranteed to fail on flat deviation functions (singular Hessian), and a second full swarm as the failure fallback — even when the 3×N control-point seed sampling has already shown the deviation is zero to machine precision. On analytic geometry (line/circle on plane/cylinder/sphere/…, i.e. most prismatic CAD) this accounts for ~18% of the total cost of a boolean operation (callgrind numbers in #1372).

This change accepts the sampled deviation and skips the optimization pipeline when every control point lies below the geometric noise floor (`Precision::SquareConfusion()` on the squared distance):

- **immediately**, when the whole composition is analytic (3D curve and pcurve of type Line/Circle/Ellipse/Hyperbola/Parabola on an elementary surface): the deviation function of such a composition is a smooth low-degree function already resolved by the control-point sampling used to seed the swarm;
- **after confirming flatness by resampling at half-step offsets** (doubling the sampling density), for all other compositions.

Sub-interval handling, the seeding itself, and every non-flat path are unchanged: any sampled squared distance above the noise floor falls through to the existing PSO + Newton pipeline. Deviations that are real but flat (e.g. a pcurve on a cylinder of a different radius) are above the floor at every sample, so they take the existing path — covered by the new gtest.

**Measurements** (single-threaded; details in #1372):
- native (gcc 13 -O2): baseplate cut with 9 rounded-box tools 59.5 → 47.4 ms (−20.3%); 5-solid fuse 28.0 → 23.2 ms (−17.1%)
- WebAssembly (Emscripten 5.0.3, Node 24): same workloads 100.9 → 80.8 ms (−19.9%) and 51.2 → 41.4 ms (−19.1%)
- `TargetFunc::Value` evaluations: 495 728 → 304 832 with the analytic gate alone; the confirmation resample removes most of the remainder
- result shapes bit-identical on both workloads (topology counts, volume, area, tolerances, `BRepCheck_Analyzer`)

**Validation:**
- DRAW test groups `boolean`, `blend`, `lowalgos`: outcome sets identical between patched and unpatched builds on the same machine (2319 OK; identical pre-existing BAD/FAILED sets) — the change does not alter any test outcome
- GTest suite: no new failures; adds `GeomLib_CheckCurveOnSurface_Test.cxx` covering a coincident analytic composition (zero deviation), a deviating analytic composition (real constant deviation must be measured, not swallowed by the fast path), and line-on-plane

Especially relevant for single-threaded consumers (WebAssembly builds), where `OSD_Parallel` cannot hide the cost of this check.

### DIFF
--- a/src/ModelingData/TKGeomBase/GTests/FILES.cmake
+++ b/src/ModelingData/TKGeomBase/GTests/FILES.cmake
@@ -40,6 +40,7 @@ set(OCCT_TKGeomBase_GTests_FILES
   GeomConvert_CompCurveToBSplineCurve_Test.cxx
   Geom2dConvert_CompCurveToBSplineCurve_Test.cxx
   GeomConvert_Test.cxx
+  GeomLib_CheckCurveOnSurface_Test.cxx
   GProp_PEquation_Test.cxx
   GProp_PGProps_Test.cxx
   Hermit_Test.cxx
--- a/src/ModelingData/TKGeomBase/GTests/GeomLib_CheckCurveOnSurface_Test.cxx
+++ b/src/ModelingData/TKGeomBase/GTests/GeomLib_CheckCurveOnSurface_Test.cxx
@@ -1,0 +1,110 @@
+// Copyright (c) 2026 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in OCCT
+// distribution for complete text of the license and disclaimer of any warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <gtest/gtest.h>
+
+#include <Adaptor3d_CurveOnSurface.hxx>
+#include <Geom2dAdaptor_Curve.hxx>
+#include <Geom2d_Line.hxx>
+#include <GeomAdaptor_Curve.hxx>
+#include <GeomAdaptor_Surface.hxx>
+#include <GeomLib_CheckCurveOnSurface.hxx>
+#include <Geom_Circle.hxx>
+#include <Geom_CylindricalSurface.hxx>
+#include <Geom_Line.hxx>
+#include <Geom_Plane.hxx>
+#include <Precision.hxx>
+#include <gp_Ax2.hxx>
+#include <gp_Ax3.hxx>
+#include <gp_Dir.hxx>
+#include <gp_Dir2d.hxx>
+#include <gp_Pnt.hxx>
+#include <gp_Pnt2d.hxx>
+
+namespace
+{
+// Circle of radius theRadius at height theHeight on a cylinder of radius
+// theCylRadius around OZ; the pcurve is the corresponding iso line in UV.
+GeomLib_CheckCurveOnSurface makeCircleOnCylinderCheck(const double theCircRadius,
+                                                      const double theCylRadius,
+                                                      const double theHeight,
+                                                      double&      theMaxDistance)
+{
+  const double aFirst = 0.0, aLast = 2.0 * M_PI;
+
+  const occ::handle<Geom_Circle> aCirc = new Geom_Circle(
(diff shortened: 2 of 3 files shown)


dpasukhi:
Dear @andymai 

andymai:
@dpasukhi CLA signed, just awaiting confirmation.

andymai:
@dpasukhi CLA approved. ID 1148

dpasukhi:
I proceed with review/improvements ;)

As our usual workflow, I will not provide you remark list, I will apply them on place as new commit inside your repo.
After my modification, your PR will be merged as Squash and Merge ;)

During today I planning to review/improve your PR.

Many thanks for your contribution and apologies for delay, looks like manager just missed 5 people in the middle of the reports ;(

dpasukhi:
I will rework the logic :( and will reduce its type coverage because the current sampling check is not safe.

Analytic geometry does not always produce a low-degree deviation function.
For example, a line pcurve on a torus can have zero deviation at every sampled point while reaching a real maximum deviation of N between samples. Half-step resampling has the same aliasing problem.

The new approach will only optimise cases such as line-line pairs and circles with exactly matching parameter frames. All other types will use the existing PSO and Newton path unchanged.

Later you can continue with the investigation in case if you will found performance is not improved as you expected.

dpasukhi (approved):

PR #1375, opened 2026-07-19
Modeling - Optimize BRepCheck_Face classification for faces with many holes

bwross:
`ClassifyWires()` currently checks every pair of a face's wires against each
other. For a plate with thousands of non-overlapping holes, that nested loop
dominates `BRepCheck_Analyzer`'s runtime, and each check was going through a
full point classification plus an iterative curve-parameter search.

This PR has been split into two commits to make it easier to review. The two
fixes that together make validating a face with many hole wires about 8x
faster:

1. 1176043c771ed6842964712bb7ee389b82bb5e68 `GeomLib_Tool::Parameter()` now uses ElCLib's closed-form
   parameter computation for analytic curve types (Line, Circle, Ellipse,
   Hyperbola, Parabola) instead of always running the iterative Extrema_ExtPC
   search. `BRepCheck_Face::ClassifyWires` calls this once per wire edge to build
   a pcurve, so it adds up fast on a face with a lot of circular holes.

2. b06b9c9dc2f7fa5d2063acf98b2269d12df91348 `BRepCheck_Face::ClassifyWires()` now precomputes a 2D bounding
   box per wire and skips the exact point-classification test whenever two
   wires' boxes don't overlap, since a point outside a wire's box can't be
   inside its finite bounded region. This is the same pre-filter pattern
   `IntersectWires()` already uses above it in the same file.

Verified against current master with a [standalone test file](https://github.com/user-attachments/files/30158374/verify_perf_fix.cpp).
Timing at 500/1000/2000 holes: 0.57s/2.26s/9.21s before, down to
0.077s/0.29s/1.11s with both fixes applied. It's about 8x faster at 2000
holes.

CLA ID is 1145.

### DIFF
--- a/src/ModelingAlgorithms/TKTopAlgo/BRepCheck/BRepCheck_Face.cxx
+++ b/src/ModelingAlgorithms/TKTopAlgo/BRepCheck/BRepCheck_Face.cxx
@@ -68,6 +68,10 @@ static bool IsInside(const TopoDS_Wire&             wir,
                      const BRepTopAdaptor_FClass2d& FClass2d,
                      const TopoDS_Face&             F);
 
+static bool ComputeWireUVBounds(const TopoDS_Face& theFace,
+                                const TopoDS_Wire& theWire,
+                                Bnd_Box2d&         theBox);
+
 static bool CheckThin(const TopoDS_Shape& w, const TopoDS_Shape& f);
 
 //=================================================================================================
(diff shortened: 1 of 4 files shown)


dpasukhi:
Dear bwross.

GeomLib_Tool::Parameter() changes are reverted.

ElCLib::Parameter() assumes the point lies on the curve and computes an inverse parameter, not the nearest parameter for a point within MaxDist. This changes existing behavior for ellipses, hyperbolas, parabolas, and ambiguous circle points.
The existing Extrema_ExtPC path already dispatches analytic curves to specialized solvers, so optimization should happen there while preserving its projection semantics.

dpasukhi (approved):

dpasukhi:
Thank you for your contribution!

Merged into IR and will be available in 8.0.1 version.

PR #1388, opened 2026-07-21
DataExchange, XCAF - AutoNamingScope guards XCAFDoc_ShapeTool::theAutoNaming

### LINKED ISSUE #1387: XCAFDoc_ShapeTool::theAutoNaming global race: RWMesh_CafReader save/modify/restore has no locking
## Summary

`XCAFDoc_ShapeTool::theAutoNaming` — a process-global `static bool` in `XCAFDoc_ShapeTool.cxx` — is
read and written with no synchronization by multiple public entry points, so two threads doing
unrelated XDE work (e.g. one importing an OBJ file while another exports a different document) can
race on it. Confirmed via ThreadSanitizer against V8_0_0_p1.

## Where

- `RWMesh_CafReader::fillDocument()` (the shared base of `RWObj_CafReader` and `RWGltf_CafReader`,
  so this is reachable from OBJ and glTF import) does a save/modify/restore dance with no lock:
  ```cpp
  const bool wasAutoNaming = XCAFDoc_ShapeTool::AutoNaming();  // read
  XCAFDoc_ShapeTool::SetAutoNaming(false);                     // write
  ...                                                          // build the document tree
  XCAFDoc_ShapeTool::SetAutoNaming(wasAutoNaming);              // restore
  ```
- `XCAFDoc_ShapeTool::addShape` reads the same flag internally (`if (theAutoNaming) { SetLabelNameByShape(...); }`),
  so any concurrent caller of `AddShape` (e.g. an OBJ/PLY export building its own document from an
  existing shape) is also racing against it.

Since `SetAutoNaming`/`AutoNaming` are public, documented API (meant to express one process-wide
setting), the fix is not simply making `theAutoNaming` `thread_local` — that would silently change
the semantics for direct application callers that set the mode on one thread and expect other
threads to observe it. The narrower, correct fix is a mutex protecting the save/modify/restore
window in `fillDocument()` (and any other internal caller that does the same dance), so overlapping
callers serialize on that critical section without changing what `theAutoNaming` *means*.

## Repro

Minimal C++ reproducer (no GUI, no Draw): N threads concurrently write a small box to its own
uniquely-named OBJ file (`RWObj_CafWriter`) then read it back (`RWObj_CafReader`) into a fresh
`TDocStd_Document`. Not a file-path collision — every thread uses its own file.

```cpp
// build+mesh independently per thread, then:
Handle(TDocStd_Document) writeDoc = new TDocStd_Document("BinXCAF");
XCAFDoc_DocumentTool::ShapeTool(writeDoc->Main())->AddShape(box);
RWObj_CafWriter writer(path.c_str());
writer.Perform(writeDoc, fileInfo, Message_ProgressRange());

Handle(TDocStd_Document) readDoc = new TDocStd_Document("BinXCAF");
RWObj_CafReader reader;
reader.SetDocument(readDoc);
reader.Perform(path.c_str(), Message_ProgressRange());
```

Full harness attached / linked below. Built OCCT as a minimal-module ThreadSanitizer instrumented
static build (`FoundationClasses`+`ModelingData`+`ModelingAlgorithms`+`DataExchange`,
`RelWithDebInfo`, `-fsanitize=thread -g`), ran with `MMGT_OPT=0 TSAN_OPTIONS="halt_on_error=0"`:

```
occt_341_stress obj_roundtrip_unique 8 25 /tmp/scratch
```

## Observed

8 threads × 25 iterations reports 9 distinct ThreadSanitizer races per run (10 threads × 200 →
17/run); every one resolves to `XCAFDoc_ShapeTool.cxx:693` (`SetAutoNaming`) /
`XCAFDoc_ShapeTool.cxx:700` (`AutoNaming`) / `XCAFDoc_ShapeTool.cxx:532` (`addShape`'s internal
read), all on the same global `theAutoNaming`.

## Environment

- OCCT V8_0_0_p1, macOS arm64 (Apple clang, AppleClang 21.0.0.21000101)
- Minimal-module static build, ThreadSanitizer (`-fsanitize=thread`)

## Suggested fix

A `static std::mutex` (or `Standard_Mutex`) guarding the read/write/read/write window in
`RWMesh_CafReader::fillDocument()` (and any sibling internal caller doing the same
save/modify/restore). Happy to submit a PR if useful — we've carried a bridge-side mitigation
(serializing all OBJ/glTF/PLY calls on our own lock) in the meantime:
https://github.com/SecondMouseAU/OCCTSwift/tree/main/Scripts/repro/341-meshcaf

gsdali:
## Summary

`XCAFDoc_ShapeTool::theAutoNaming` is a process-global `static bool` (documented as such — "This
setting is global; it cannot be made a member function") that three independent call sites each
save/modify/restore around their own work with no locking:

- `RWMesh_CafReader::fillDocument()` — shared base of `RWObj_CafReader` and `RWGltf_CafReader`, so
  reachable from both OBJ and glTF import
- `RWGltf_CafReader::fillDocument()` — a *separate*, near-duplicate override, not a call into the
  base class's version
- `XCAFDoc_Editor::Expand()` — additionally recurses into itself while the save/restore is in flight

`XCAFDoc_ShapeTool::AddShape`/`addShape` reads the same flag internally to decide whether to
auto-generate a name.

ThreadSanitizer on a concurrent-OBJ-import stress (8-10 threads, each writing/reading its own
uniquely-named file — not a file-path collision) reports 9-17 races per run on `theAutoNaming`,
resolving to two distinct problems:

1. Two threads' save/modify/restore critical sections can interleave, so one thread's restore
   stomps another's still-in-progress override — a logical bug (wrong auto-naming mode active for
   part of a build), independent of memory safety.
2. `theAutoNaming` itself is a plain `bool`, read and written with no synchronization from *any*
   caller — including ordinary `AddShape` calls made outside all three save/restore sites above — a
   genuine data race on every access.

## Fix

Two parts, both needed:

1. `XCAFDoc_ShapeTool::AutoNamingScope` — a new RAII helper (`AutoNamingScope(bool)` constructor /
   destructor) that holds a `std::recursive_mutex` for its *entire* lifetime, not just around the
   individual get/set calls, so two threads' save-modify-restore sequences serialize instead of
   interleaving. Recursive because `XCAFDoc_Editor::Expand()` reenters it on the same thread. All
   three call sites now use it; `Expand()`'s two duplicate manual-restore-before-return sites
   collapse into one destructor-driven restore that fires on every exit path.
2. `theAutoNaming` is now `std::atomic<bool>` instead of a plain `bool`, so every access anywhere in
   the file — including `AddShape`'s internal read, which participates in none of the three scoped
   sections — is well-defined, closing the gap the mutex alone doesn't reach.

## Testing

Minimal-module ThreadSanitizer build (`FoundationClasses`+`ModelingData`+`ModelingAlgorithms`+
`DataExchange`, `RelWithDebInfo`, `-fsanitize=thread`). A pure-C++ harness runs N threads each
building a box, meshing it, writing it to its own uniquely-named `.obj` file (`RWObj_CafWriter`),
and reading it back (`RWObj_CafReader`) into a fresh `TDocStd_Document`.

- Before: 10 threads × 200 iterations reports 9-17 `theAutoNaming` races per run (4 runs).
- After: **0** `theAutoNaming` races across the same stress, 4 separate runs — only an unrelated,
  pre-existing `Message_PrinterOStream`/`std::cout` console-write race remains (OCCT's default
  messenger has no internal locking; out of scope for this PR).
- Zero regression on unrelated concurrent-shape-creation and independent-meshing scenarios.
- `RWGltf_CafReader`'s copy of the fix compiles cleanly (mechanically identical to the
  `RWMesh_CafReader` path that *was* exercised under TSan) but wasn't run under TSan directly in this
  session — the minimal-module build excludes `TKDEGLTF` (needs RapidJSON, disabled for build speed).

No behavior change for the existing (documented) semantics of `SetAutoNaming`/`AutoNaming` as a
single process-wide setting — this only serializes the internal save/modify/restore idiom and makes
the underlying storage well-defined.

Fixes #1387.

### DIFF
--- a/src/DataExchange/TKDEGLTF/RWGltf/RWGltf_CafReader.cxx
+++ b/src/DataExchange/TKDEGLTF/RWGltf/RWGltf_CafReader.cxx
@@ -449,20 +449,18 @@ void RWGltf_CafReader::fillDocument()
     Message::SendWarning("Warning: Length unit of document not equal to the system length unit");
   }
 
-  const bool wasAutoNaming = XCAFDoc_ShapeTool::AutoNaming();
-  XCAFDoc_ShapeTool::SetAutoNaming(false);
   const TCollection_AsciiString aRootName; // = generateRootName (theFile);
   CafDocumentTools              aTools;
   aTools.ShapeTool       = XCAFDoc_DocumentTool::ShapeTool(myXdeDoc->Main());
   aTools.ColorTool       = XCAFDoc_DocumentTool::ColorTool(myXdeDoc->Main());
   aTools.VisMaterialTool = XCAFDoc_DocumentTool::VisMaterialTool(myXdeDoc->Main());
+  XCAFDoc_ShapeTool::OwnAutoNamingScope anAutoNamingScope(aTools.ShapeTool, false);
   for (NCollection_Sequence<TopoDS_Shape>::Iterator aRootIter(myRootShapes); aRootIter.More();
        aRootIter.Next())
   {
     addShapeIntoDoc(aTools, aRootIter.Value(), TDF_Label(), aRootName);
   }
   XCAFDoc_DocumentTool::ShapeTool(myXdeDoc->Main())->UpdateAssemblies();
-  XCAFDoc_ShapeTool::SetAutoNaming(wasAutoNaming);
 }
 
 //=================================================================================================
--- a/src/DataExchange/TKRWMesh/RWMesh/RWMesh_CafReader.cxx
+++ b/src/DataExchange/TKRWMesh/RWMesh/RWMesh_CafReader.cxx
@@ -178,20 +178,18 @@ void RWMesh_CafReader::fillDocument()
     Message::SendWarning("Warning: Length unit of document not equal to the system length unit");
   }
 
-  const bool wasAutoNaming = XCAFDoc_ShapeTool::AutoNaming();
-  XCAFDoc_ShapeTool::SetAutoNaming(false);
   const TCollection_AsciiString aRootName; // = generateRootName (theFile);
   CafDocumentTools              aTools;
   aTools.ShapeTool       = XCAFDoc_DocumentTool::ShapeTool(myXdeDoc->Main());
   aTools.ColorTool       = XCAFDoc_DocumentTool::ColorTool(myXdeDoc->Main());
   aTools.VisMaterialTool = XCAFDoc_DocumentTool::VisMaterialTool(myXdeDoc->Main());
+  XCAFDoc_ShapeTool::OwnAutoNamingScope anAutoNamingScope(aTools.ShapeTool, false);
   for (NCollection_Sequence<TopoDS_Shape>::Iterator aRootIter(myRootShapes); aRootIter.More();
        aRootIter.Next())
   {
     addShapeIntoDoc(aTools, aRootIter.Value(), TDF_Label(), aRootName);
   }
   XCAFDoc_DocumentTool::ShapeTool(myXdeDoc->Main())->UpdateAssemblies();
-  XCAFDoc_ShapeTool::SetAutoNaming(wasAutoNaming);
 }
 
 //=================================================================================================
--- a/src/DataExchange/TKXCAF/XCAFDoc/XCAFDoc_Editor.cxx
+++ b/src/DataExchange/TKXCAF/XCAFDoc/XCAFDoc_Editor.cxx
@@ -66,9 +66,8 @@ bool XCAFDoc_Editor::Expand(const TDF_Label& theDoc,
   {
     return false;
   }
-  occ::handle<XCAFDoc_ShapeTool> aShapeTool   = XCAFDoc_DocumentTool::ShapeTool(theDoc);
-  bool                           isAutoNaming = aShapeTool->AutoNaming();
-  aShapeTool->SetAutoNaming(false);
+  occ::handle<XCAFDoc_ShapeTool>        aShapeTool = XCAFDoc_DocumentTool::ShapeTool(theDoc);
+  XCAFDoc_ShapeTool::OwnAutoNamingScope anAutoNamingScope(aShapeTool, false);
 
   TDF_Label aCompoundPartL = theShape;
   if (aShapeTool->IsReference(theShape))
@@ -139,10 +138,8 @@ bool XCAFDoc_Editor::Expand(const TDF_Label& theDoc,
         }
       }
     }
-    aShapeTool->SetAutoNaming(isAutoNaming);
     return true;
   }
-  aShapeTool->SetAutoNaming(isAutoNaming);
   return false;
 }
 
--- a/src/DataExchange/TKXCAF/XCAFDoc/XCAFDoc_ShapeTool.cxx
+++ b/src/DataExchange/TKXCAF/XCAFDoc/XCAFDoc_ShapeTool.cxx
@@ -15,6 +15,8 @@
 
 #include <XCAFDoc_ShapeTool.hxx>
 
+#include <atomic>
+
 #include <BRep_Builder.hxx>
 #include <gp_Pnt.hxx>
 #include <gp_Trsf.hxx>
@@ -50,7 +52,7 @@
 
 IMPLEMENT_DERIVED_ATTRIBUTE_WITH_TYPE(XCAFDoc_ShapeTool, TDataStd_GenericEmpty, "xcaf", "ShapeTool")
 
-static bool theAutoNaming = true;
+static std::atomic<bool> theAutoNaming{true};
 
 // attribute methods //////////////////////////////////////////////////
 
@@ -81,6 +83,7 @@ occ::handle<XCAFDoc_ShapeTool> XCAFDoc_ShapeTool::Set(const TDF_Label& L)
 XCAFDoc_ShapeTool::XCAFDoc_ShapeTool()
 {
   hasSimpleShapes = false;
+  myOwnAutonaming = -1;
 }
 
 //=================================================================================================
@@ -453,7 +456,7 @@ void XCAFDoc_ShapeTool::MakeReference(const TDF_Label&       L,
   refNode->Remove(); // abv: fix against bug in TreeNode::Append()
   mainNode->Append(refNode);
 
-  if (theAutoNaming)
+  if (OwnAutoNaming())
   {
     SetLabelNameByLink(L);
   }
@@ -529,7 +532,7 @@ TDF_Label XCAFDoc_ShapeTool::addShape(const TopoDS_Shape& S, const bool makeAsse
   //  }
   A->SetShape(S);
 
-  if (theAutoNaming)
+  if (OwnAutoNaming())
   {
     SetLabelNameByShape(ShapeLabel);
   }
@@ -540,7 +543,7 @@ TDF_Label XCAFDoc_ShapeTool::addShape(const TopoDS_Shape& S, const bool makeAsse
     // mark assembly by assigning UAttribute
     occ::handle<TDataStd_UAttribute> Uattr;
     Uattr = TDataStd_UAttribute::Set(ShapeLabel, XCAFDoc::AssemblyGUID());
-    if (theAutoNaming)
+    if (OwnAutoNaming())
     {
       TDataStd_Name::Set(ShapeLabel, TCollection_ExtendedString("ASSEMBLY"));
     }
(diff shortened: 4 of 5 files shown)


gkv311:
@gsdali 

I think that a mutex is not the right tool here and I don't quite understand how current patch fixes concurrency issue as `theAutoNaming` usage remains unprotected within `XCAFDoc_ShapeTool.cpp` itself.

What I would suggest is to add local property overriding global value as a `XCAFDoc_ShapeTool` class field and use it in specific places. I have prototyped the patch for an older OCCT - fill free to borrow the idea for your patch:

```diff
---------------------- src/XCAFDoc/XCAFDoc_ShapeTool.hxx
   Standard_EXPORT static Standard_Boolean AutoNaming();
-  
+
+  //! Return autonaming flag for this tool instance.
+  Standard_Boolean OwnAutoNaming() const
+  {
+    return myOwnAutonaming == -1 ? XCAFDoc_ShapeTool::AutoNaming() : myOwnAutonaming == 1;
+  }
+
+  //! Set autonaming flag for this tool instance.
+  void SetOwnAutoNaming (Standard_Boolean theOwnFlag)
+  {
+    myOwnAutonaming = theOwnFlag ? 1 : 0;
+  }
+
+  //! Reset autonaming flag for this tool instance to global state.
+  void UnsetOwnAutoNaming()
+  {
+    myOwnAutonaming = -1;
+  }
...
   //! Makes a shape on label L to be a reference to shape refL
   //! with location loc
-  Standard_EXPORT static void MakeReference (const TDF_Label& L, const TDF_Label& refL, const TopLoc_Location& loc);
+  Standard_EXPORT void MakeReference (const TDF_Label& L, const TDF_Label& refL, const TopLoc_Location& loc);
...
-  Standard_Boolean hasSimpleShapes;
-
+  Standard_Boolean hasSimpleShapes = false;
+  //! Define own autonaming property; -1 by default which means to use global flag
+  Standard_Integer myOwnAutonaming = -1;
 
 };
...
--------------------- src/XCAFDoc/XCAFDoc_ShapeTool.cxx ----------------------
index 3b36233880..b3357c9d65 100644
@@ -454,7 +454,7 @@ void XCAFDoc_ShapeTool::MakeReference (const TDF_Label &L,
   refNode->Remove(); // abv: fix against bug in TreeNode::Append()
   mainNode->Append(refNode);
 
-  if (theAutoNaming)
+  if (OwnAutoNaming())
     SetLabelNameByLink(L);
 }
 
@@ -529,7 +529,7 @@ TDF_Label XCAFDoc_ShapeTool::addShape (const TopoDS_Shape& S, const Standard_Boo
 //  }
   A->SetShape(S);
   
-  if (theAutoNaming)
+  if (OwnAutoNaming())
     SetLabelNameByShape(ShapeLabel);
 
   // if shape is Compound and flag is set, create assembly
@@ -537,7 +537,7 @@ TDF_Label XCAFDoc_ShapeTool::addShape (const TopoDS_Shape& S, const Standard_Boo
     // mark assembly by assigning UAttribute
     Handle(TDataStd_UAttribute) Uattr;
     Uattr = TDataStd_UAttribute::Set ( ShapeLabel, XCAFDoc::AssemblyGUID() );
-    if (theAutoNaming)
+    if (OwnAutoNaming())
       TDataStd_Name::Set(ShapeLabel, TCollection_ExtendedString("ASSEMBLY"));
 
     // iterate on components
@@ -1564,7 +1564,7 @@ Standard_Boolean XCAFDoc_ShapeTool::SetSHUO (const TDF_LabelSequence& labels,
   
   TDF_TagSource aTag;
   TDF_Label UpperSubL = aTag.NewChild( labels( 1 ) );
-  if (theAutoNaming) {
+  if (OwnAutoNaming()) {
     TCollection_ExtendedString Entry("SHUO");
     TDataStd_Name::Set(UpperSubL, TCollection_ExtendedString( Entry ));
   }
@@ -1575,7 +1575,7 @@ Standard_Boolean XCAFDoc_ShapeTool::SetSHUO (const TDF_LabelSequence& labels,
   // add other next_usage occurrences.
   for (i = 2; i <= labels.Length(); i++) {
     TDF_Label NextSubL = aTag.NewChild( labels( i ) );
-    if (theAutoNaming) {
+    if (OwnAutoNaming()) {

------------------------ src/XCAFDoc/XCAFDoc_Editor.cxx ------------------------
index 6897238bc3..b18adf6217 100644
@@ -68,8 +68,7 @@ Standard_Boolean XCAFDoc_Editor::Expand (const TDF_Label& theDoc,
     return Standard_False;
   }
   Handle(XCAFDoc_ShapeTool) aShapeTool = XCAFDoc_DocumentTool::ShapeTool(theDoc);
-  Standard_Boolean isAutoNaming = aShapeTool->AutoNaming();
-  aShapeTool->SetAutoNaming(Standard_False);
+  aShapeTool->SetOwnAutoNaming(false);
 
   TDF_Label aCompoundPartL = theShape;
   if (aShapeTool->IsReference(theShape))
@@ -137,10 +136,10 @@ Standard_Boolean XCAFDoc_Editor::Expand (const TDF_Label& theDoc,
         }
       }
     }
-    aShapeTool->SetAutoNaming(isAutoNaming);
+    aShapeTool->UnsetOwnAutoNaming();
     return Standard_True;
   }
-  aShapeTool->SetAutoNaming(isAutoNaming);
+  aShapeTool->UnsetOwnAutoNaming();
   return Standard_False;
 }
----------------------- src/RWMesh/RWMesh_CafReader.cxx -----------------------
index 4164e4496f..d021a51043 100644
@@ -178,11 +178,10 @@ void RWMesh_CafReader::fillDocument()
     Message::SendWarning("Warning: Length unit of document not equal to the system length unit");
   }
 
-  const Standard_Boolean wasAutoNaming = XCAFDoc_ShapeTool::AutoNaming();
-  XCAFDoc_ShapeTool::SetAutoNaming (Standard_False);
   const TCollection_AsciiString aRootName; // = generateRootName (theFile);
   CafDocumentTools aTools;
   aTools.ShapeTool = XCAFDoc_DocumentTool::ShapeTool (myXdeDoc->Main());
+  aTools.ShapeTool->SetOwnAutoNaming (false);
   aTools.ColorTool = XCAFDoc_DocumentTool::ColorTool (myXdeDoc->Main());
   aTools.VisMaterialTool = XCAFDoc_DocumentTool::VisMaterialTool (myXdeDoc->Main());
   for (TopTools_SequenceOfShape::Iterator aRootIter (myRootShapes); aRootIter.More(); aRootIter.Next())
@@ -190,7 +189,7 @@ void RWMesh_CafReader::fillDocument()
     addShapeIntoDoc (aTools, aRootIter.Value(), TDF_Label(), aRootName);
   }
   XCAFDoc_DocumentTool::ShapeTool (myXdeDoc->Main())->UpdateAssemblies();
-  XCAFDoc_ShapeTool::SetAutoNaming (wasAutoNaming);
+  aTools.ShapeTool->UnsetOwnAutoNaming();
 }
```

In this way, the API with the global setting will remain working, while allowing to disable auto-naming locally within specific algorithms.

gsdali:
@gkv311 Thank you for the review — and you're right, a mutex isn't the right tool here.

We'd used a mutex-guarded save/restore successfully for a couple of similar unsynchronized-global-state issues elsewhere in this codebase, so that was the pattern we defaulted to without stopping to ask whether the state should have been global in the first place. In hindsight this is a clearly better fix: `theAutoNaming` was only ever being overridden by three call sites that each want to suppress naming for *their own* document build, not change a process-wide setting — so the override belongs on `XCAFDoc_ShapeTool` (already one instance per document), not on a shared flag. That's the original intent the three call sites were reaching for anyway; the global was the wrong tool even before concurrency entered the picture.

Pushed an update using your suggested design: `myOwnAutonaming` (tri-state, -1 inherits the process-wide default) plus `OwnAutoNaming()`/`SetOwnAutoNaming()`/`UnsetOwnAutoNaming()`, with `AddShape`/`MakeReference`/`SetSHUO` reading `OwnAutoNaming()` instead of the flag directly. One thing worth flagging since it wasn't obvious from the one-line sketch: `XCAFDoc_Editor::Expand()` recurses into itself on the same document, so a bare `SetOwnAutoNaming()`/`UnsetOwnAutoNaming()` pair at entry/exit would clobber an outer call's override mid-recursion (the inner call's unconditional "reset to inherit global" would win over the outer call's still-active override). `OwnAutoNamingScope` does a proper save/restore of whatever override state the instance had on entry — same shape as the RAII scope this replaces, just scoped to the instance instead of a process-wide flag, so nesting composes correctly with no lock needed either way.

Re-verified with the same ThreadSanitizer stress as before (10 threads x 200 iterations) — zero `theAutoNaming` races, same result as the previous fix. Also added a stress scenario for the specific gap you flagged: half the threads locally override via `OwnAutoNamingScope` on their own document while the other half do plain unscoped `AddShape()` on independent documents relying on the process-wide default, concurrently — 3000 operations, zero instances of the unscoped threads observing another thread's override.

gkv311:
@gsdali indeed, thanks for pointing out - haven't noticed that.

> One thing worth flagging since it wasn't obvious from the one-line sketch: XCAFDoc_Editor::Expand() recurses into itself on the same document

PR #1399, opened 2026-07-24
Foundation Classes - Fix Resource_Manager::Debug and Storage_Schema current-data data races

### LINKED ISSUE #1398: Resource_Manager::Debug and Storage_Schema::ICurrentData() race under legitimate multi-instance TDocStd_Application usage
### Description

Found while migrating a consumer (OCCTSwift) away from `XCAFApp_Application::GetApplication()`
per the guidance in #1396 — a private `TDocStd_Application` per caller, never shared across
threads, is otherwise exactly the "no shared mutable state" usage pattern that should be free of
the #1387/#1389/#1393/#1396 class of races. It mostly is, but two additional, previously
unreachable races surface once `Resources()`/persistence-format registration and document
open/save genuinely run concurrently on *independent* application instances:

**1. `Resource_Manager::Resource_Manager(const char*, bool)`** (`Resource_Manager.cxx:107-109`)
writes a file-scope global on every construction, unsynchronized:

```cpp
Resource_Manager::Resource_Manager(const char* const aName, const bool Verbose)
    : myName(aName), myVerbose(Verbose), myInitialized(false)
{
  OSD_Environment envDebug("ResourceDebug");
  Debug = (!envDebug.Value().IsEmpty());   // <- `static bool Debug;` at file scope, line 54
  ...
```

Reached via `TDocStd_Application::Resources()` (called from `DefineFormat()`) constructing its
lazily-initialized `Resource_Manager`. `Resources()`'s own lazy-init is already mutex-guarded
per-instance (from the #1389/#1390 fix), but that only protects one `TDocStd_Application`
instance calling `Resources()` twice — it does nothing when N *different* instances each
construct their own `Resource_Manager` concurrently, which all write the same global `Debug`.

**2. `Storage_Schema::ICurrentData()`** (`Storage_Schema.cxx:802-805`) is a function-local static
`Handle(Storage_Data)` — thread-safe to *initialize* (C++11 magic statics), but every
`Storage_Schema` instance mutates the handle it returns with no synchronization at all:

```cpp
occ::handle<Storage_Data>& Storage_Schema::ICurrentData()
{
  static occ::handle<Storage_Data> _Storage_CData;
  return _Storage_CData;
}
```

`Storage_Schema::Storage_Schema()`'s constructor immediately calls `Clear()`, which does
`Storage_Schema::ICurrentData().Nullify()` — so constructing a new `Storage_Schema` on one thread
can null out the handle another thread's in-flight (de)serialization is concurrently reading via
`ICurrentData()->InternalData()->...`. Reached via `PCDM_ReadWriter_1::ReadDocumentVersion` during
`TDocStd_Application::Open()`.

### Why this was never caught before

Both are latent bugs, but neither had ever been exercised concurrently: with every document
sharing one `XCAFApp_Application::GetApplication()` instance (the pattern #1396 itself questions),
`Resources()`'s per-instance lazy-init effectively made `Resource_Manager` construction — and by
extension `Storage_Schema` usage during Open/SaveAs — happen essentially once, serially, for the
whole process. Moving to a private application instance per caller (as suggested in #1396) is what
first makes these concurrent.

### Expected Behavior

Independent `TDocStd_Application` instances, each used by only one thread at a time, performing
`DefineFormat`/`Open`/`SaveAs` concurrently should not race on process-global state.

### Actual Behavior

ThreadSanitizer (minimal-module build: FoundationClasses+ModelingData+ModelingAlgorithms+
DataExchange) reports 16 race warnings across both sites (8 threads x 50 rounds, each round: a
brand-new private app, `DefineFormat`, `SaveAs`, `Close`, then a *separate* brand-new private app,
`Open`, verify, `Close` — no two threads ever touch the same `TDocStd_Application` instance).

### Sample Code

Full reproducer: https://github.com/SecondMouseAU/OCCTSwift/blob/main/Scripts/repro/371-getapplication-singleton-elimination/occt_371_private_app.cpp

### Operating System

macOS

### Compiler

Clang

### Bitness

64-bit

### OCCT Version

latest

### Additional Files

Full writeup: https://github.com/SecondMouseAU/OCCTSwift/tree/main/Scripts/repro/371-getapplication-singleton-elimination

gsdali:
## Description

Fixes #1398.

`Resource_Manager::Resource_Manager(const char*, bool)` (`Resource_Manager.cxx:109`) writes a
file-scope `static bool Debug` on every construction with zero synchronization. Two
`Resource_Manager` instances constructed concurrently on different threads (e.g. two independent
`TDocStd_Application` instances each lazily building their own via `Resources()`/`DefineFormat()`)
race on this write. Fixed by making `Debug` a `std::atomic<bool>`: it is a plain process-wide
debug-logging flag, not per-instance intent, so an atomic is sufficient here.

`Storage_Schema::ICurrentData()` (`Storage_Schema.cxx:802`) is a function-local static
`Handle(Storage_Data)`, shared by every `Storage_Schema` in the process and mutated with no
synchronization anywhere. `Write()` sets it for the duration of one store, and *any*
`Storage_Schema` construction calls `Clear()` -> `ICurrentData().Nullify()` in its constructor,
including throwaway instances built only to read header info
(`PCDM_ReadWriter_1::ReadReferenceCounter`, `ReadReferences` and `ReadDocumentVersion`, all reached
unconditionally from a plain `TDocStd_Application::Open()` via `CDF_Application::Retrieve`). One
instance's construction can therefore null out a *different*, unrelated instance's in-flight
`Write()` on another thread, or race a concurrent load's own throwaway construction.

**Fixed by removing the global rather than guarding it**, per @gkv311's review below: the handle
becomes a `mutable Handle(Storage_Data) myCurrentData` field on `Storage_Schema`, so there is no
shared state left to synchronize. `Write()` assigns the field, `Clear()` nullifies its own, and
`HasTypeBinding()`/`BindType()`/`TypeBinding()`/`AddPersistent()`/`PersistentToAdd()` read it
through `*this`. The private statics `ICurrentData()` and `ISetCurrentData()` are removed.
`mutable` because all of those methods are `const`.

Nothing in the tree relies on the state being process-wide:

- Every `Storage_Schema` is constructed locally by its caller and used only there
  (`PCDM_StorageDriver::Write`, and `PCDM_ReadWriter_1` at three sites). None is cached or shared.
- `Storage_CallBack::Add`/`Write`/`Read` all take the driving schema as an argument, so the
  `BindType`/`AddPersistent`/`PersistentToAdd` calls made from per-type callbacks during a `Write()`
  run against that same instance.
- The other `Storage_Schema` users in the tree (`BinLDrivers`, `XmlMDF`, `StdLDrivers`) use only the
  statics `CheckTypeMigration()` and `ICreationDate()`, neither of which reads the current data.
- `ICurrentData()` and `ISetCurrentData()` were both `private` with no reference outside the class,
  so removing them is not an API change.

The same commit drops the `static` from `AddPersistent()`'s `TCollection_AsciiString aTypeName`, a
second unsynchronized process-wide object in the same class. It is assigned from the `tName`
argument and read two lines later, so `static` only ever saved an allocation, and it was previously
covered only incidentally by the mutex this revision removes.

Both races are invisible under the traditional single shared `XCAFApp_Application` /
`TDocStd_Application` usage pattern, because there `Resources()`'s own lazy-init mutex happens to
serialize `Resource_Manager`/`Storage_Schema` construction down to "runs once, ever, for the whole
process." They only become reachable once an application is constructed privately per caller or
thread (the pattern this project's own `TDocStd_Application` documentation recommends over the
`XCAFApp_Application::GetApplication()` singleton) and multiple independent instances are built
concurrently. See the linked issue for the full TSan writeup and a standalone reproducer.

No public API signature changes. The revision is a net reduction: 15 insertions, 31 deletions.

## Type of change

## How Has This Been Tested?

ThreadSanitizer (minimal-module build: FoundationClasses + ModelingData + ModelingAlgorithms +
DataExchange, `RelWithDebInfo`, `-fsanitize=thread -g`), each round building a brand-new private
`TDocStd_Application`, defining formats, populating and saving a document, closing it, then a
**separate** brand-new private application opening, verifying and closing it. No two threads ever
share an application or schema instance.

- Before the fix: 13 TSan race warnings (2x `Resource_Manager::Resource_Manager`, 11x
  `Storage_Schema::Storage_Schema`'s `Clear()` call) at 8 threads x 30 rounds, 0 functional
  failures.
- The mutex revision: 0 races across 4 runs (8x30, 8x50, 10x60, 8x40), 0 functional failures.
- This per-instance revision: 0 races across 8x50, 8x30 and 10x60, 0 save, load or verify failures.
  Re-run after the `aTypeName` change with the same result.

Also run against the wider concurrency suite that the downstream project maintains for this kernel
(10 scenarios covering the fixes from #1374, #1388, #1390, #1394 and #1397 plus this one): all
clean, no regression. Full downstream test suite (4666 tests) clean against a production build of
the patched kernel.

Reproducer and full writeup:
https://github.com/SecondMouseAU/OCCTSwift/tree/main/Scripts/repro/374-resource-manager-storage-schema-race

## Checklist:

- [x] My code follows the code style of this project (`clang-format --dry-run --Werror` clean on
      every touched file)
- [x] My change requires a change to the documentation
  - [ ] I have updated the documentation accordingly (no user-facing documentation covers this
        internal-only change)

### DIFF
--- a/src/FoundationClasses/TKernel/Resource/Resource_Manager.cxx
+++ b/src/FoundationClasses/TKernel/Resource/Resource_Manager.cxx
@@ -31,6 +31,7 @@
 #include <NCollection_Array1.hxx>
 
 #include <algorithm>
+#include <atomic>
 #include <cerrno>
 
 IMPLEMENT_STANDARD_RTTIEXT(Resource_Manager, Standard_Transient)
@@ -51,7 +52,7 @@ static Resource_KindOfLine WhatKindOfLine(OSD_File&                aFile,
 
 static int GetLine(OSD_File& aFile, TCollection_AsciiString& aLine);
 
-static bool Debug;
+static std::atomic<bool> Debug(false);
 
 //=================================================================================================
 
--- a/src/FoundationClasses/TKernel/Storage/Storage_Schema.cxx
+++ b/src/FoundationClasses/TKernel/Storage/Storage_Schema.cxx
@@ -318,7 +318,7 @@ void Storage_Schema::Write(const occ::handle<Storage_BaseDriver>& theDriver,
   occ::handle<Standard_Persistent>                              p;
   occ::handle<NCollection_HSequence<occ::handle<Storage_Root>>> plist;
   TCollection_AsciiString                                       errorContext("AddPersistent");
-  Storage_Schema::ISetCurrentData(aData);
+  myCurrentData = aData;
 
   occ::handle<Storage_InternalData> iData = aData->InternalData();
 
@@ -614,8 +614,8 @@ void Storage_Schema::BindType(const TCollection_AsciiString&       aTypeName,
 {
   if (!HasTypeBinding(aTypeName))
   {
-    occ::handle<Storage_InternalData>  iData = Storage_Schema::ICurrentData()->InternalData();
-    occ::handle<Storage_TypeData>      tData = Storage_Schema::ICurrentData()->TypeData();
+    occ::handle<Storage_InternalData>  iData = myCurrentData->InternalData();
+    occ::handle<Storage_TypeData>      tData = myCurrentData->TypeData();
     occ::handle<Storage_TypedCallBack> c     = new Storage_TypedCallBack(aTypeName, aCallBack);
 
     tData->AddType(aTypeName, iData->myTypeId);
@@ -633,7 +633,7 @@ occ::handle<Storage_CallBack> Storage_Schema::TypeBinding(
 
   if (HasTypeBinding(aTypeName))
   {
-    occ::handle<Storage_InternalData> iData = Storage_Schema::ICurrentData()->InternalData();
+    occ::handle<Storage_InternalData> iData = myCurrentData->InternalData();
 
     result = iData->myTypeBinding.Find(aTypeName)->CallBack();
   }
@@ -650,14 +650,13 @@ bool Storage_Schema::AddPersistent(const occ::handle<Standard_Persistent>& sp,
 
   if (!sp.IsNull())
   {
-    occ::handle<Storage_InternalData> iData = Storage_Schema::ICurrentData()->InternalData();
+    occ::handle<Storage_InternalData> iData = myCurrentData->InternalData();
 
     if (sp->_typenum == 0)
     {
-      int                            aTypenum;
-      static TCollection_AsciiString aTypeName;
-      aTypeName                           = tName;
-      occ::handle<Storage_TypeData> tData = Storage_Schema::ICurrentData()->TypeData();
+      int                           aTypenum;
+      TCollection_AsciiString       aTypeName(tName);
+      occ::handle<Storage_TypeData> tData = myCurrentData->TypeData();
 
       aTypenum = iData->myTypeBinding.Find(aTypeName)->Index();
 
@@ -679,7 +678,7 @@ bool Storage_Schema::PersistentToAdd(const occ::handle<Standard_Persistent>& sp)
 
   if (!sp.IsNull())
   {
-    occ::handle<Storage_InternalData> di = Storage_Schema::ICurrentData()->InternalData();
+    occ::handle<Storage_InternalData> di = myCurrentData->InternalData();
 
     if (sp->_typenum == 0 && sp->_refnum != -1)
     {
@@ -696,7 +695,7 @@ bool Storage_Schema::PersistentToAdd(const occ::handle<Standard_Persistent>& sp)
 
 void Storage_Schema::Clear() const
 {
-  Storage_Schema::ICurrentData().Nullify();
+  myCurrentData.Nullify();
 }
 
 #ifdef DATATYPE_MIGRATION
@@ -789,21 +788,6 @@ bool Storage_Schema::CheckTypeMigration(const TCollection_AsciiString& oldName,
   return aMigration;
 }
 #endif
-
-//=================================================================================================
-
-void Storage_Schema::ISetCurrentData(const occ::handle<Storage_Data>& dData)
-{
-  Storage_Schema::ICurrentData() = dData;
-}
-
-//=================================================================================================
-
-occ::handle<Storage_Data>& Storage_Schema::ICurrentData()
-{
-  static occ::handle<Storage_Data> _Storage_CData;
-  return _Storage_CData;
-}
 
 #define SLENGTH 80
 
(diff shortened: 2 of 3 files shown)


gkv311:
@gsdali I think `Storage_Schema::ICurrentData` could be moved to `Storage_Schema` as a `mutable Handle(Storage_Data) myCurrentData;` class field to get rid of this global variable.

gsdali:
Done, and thanks for the pointer: `ICurrentData()`'s global is gone and `Storage_Schema` now carries `mutable Handle(Storage_Data) myCurrentData` instead. `ICurrentDataMutex()` and every `lock_guard` went with it, so the revision is a net 15 insertions / 31 deletions. Force-pushed onto the same branch.

I checked the assumptions the field relies on before changing it:

- every `Storage_Schema` in the tree is constructed locally by its caller and used only there (`PCDM_StorageDriver::Write`, and `PCDM_ReadWriter_1` at three sites); none is cached or shared;
- `Storage_CallBack::Add`/`Write`/`Read` all take the driving schema as an argument, so the `BindType`/`AddPersistent`/`PersistentToAdd` calls that per-type callbacks make during a `Write()` run against that same instance;
- `ICurrentData()` and `ISetCurrentData()` were both `private` with no reference outside the class, so removing them is not an API change;
- the remaining `Storage_Schema` users (`BinLDrivers`, `XmlMDF`, `StdLDrivers`) touch only the statics `CheckTypeMigration()` and `ICreationDate()`, neither of which reads the current data.

It also fixes strictly more than the mutex did. Under the lock, a throwaway schema built by `PCDM_ReadWriter_1` during an unrelated `Open()` still nullified an in-flight `Write()`'s current data, it just did so without a data race. With the field it cannot reach another instance's data at all.

One extra change rode along in the same commit, because removing the lock re-exposes it: `AddPersistent()` had a `static TCollection_AsciiString aTypeName`, a second unsynchronized process-wide object in the same class. It is assigned from the `tName` argument and read two lines later, so the `static` only ever saved an allocation, and it is now a local. Glad to move it to its own commit if you would rather keep this one to the field change alone.

ThreadSanitizer: 0 races at 8x50, 8x30 and 10x60 with the field, matching what the mutex revision gave, against 13 race warnings plus an abort on the unpatched kernel at 8x30. Re-run after the `aTypeName` change with the same result. The wider concurrency suite that the downstream project keeps for this kernel (10 scenarios, covering #1374, #1388, #1390, #1394, #1397 and this one) is clean, and a production build of the patched kernel passes the downstream test suite.

On whether to split: I have kept `Resource_Manager` and `Storage_Schema` in one PR, since they share one issue (#1398), one reproducer, and one precondition, in that both only become reachable once applications are constructed privately per caller rather than through the shared singleton. Now that they use two different techniques I am happy to split them into two PRs if you prefer; just say which way you want it.

PR #1400, opened 2026-07-27
Modeling - Optimize exact face classification fallback

dpasukhi:
- Add a thread-safe lazy grid to CSLib_Class2d for constant-time
  classification away from polygon boundaries while preserving the exact
  tolerance path for boundary cells.
- Normalize coordinates and explicit tolerances robustly, handle invalid and
  extreme input values, and preserve cache state correctly across copy and
  move operations.
- Cache face wires, edge occurrences, pcurves, parameter ranges, and optional
  2D bounding boxes in BRepClass_FaceExplorer to avoid repeated topology and
  geometry traversal during exact classification.
- Introduce explicit cached-geometry and bounding-box states in
  BRepClass_Edge, invalidate derived data on topology changes, and isolate
  bounding-box construction failures per edge.
- Reuse a lazily constructed exact face explorer in
  BRepTopAdaptor_FClass2d, cache periodic surface properties, and serialize
  access to mutable exact-classification traversal state.
- Replace node-based classifier and point sequences with contiguous linear
  vectors, remove intermediate polygon copies, and make tighter wire
  discretization transactional.
- Consolidate Perform() and TestOnRestriction() through a shared
  classification path while preserving their boundary and periodic
  recadrement semantics.


dpasukhi:
The original ticket: https://github.com/FreeCAD/FreeCAD/issues/19329

PR #1410, opened 2026-07-29
Shape Healing - create a ReShape context when none is set in ShapeFix_ComposeShell and ShapeUpgrade_WireDivide

### LINKED ISSUE #1409: Shape Healing - ShapeFix_ComposeShell::Perform/SplitEdges and ShapeUpgrade_WireDivide::Perform SIGSEGV on a null ReShape context
### Description

`ShapeFix_ComposeShell::Perform()`, `ShapeFix_ComposeShell::SplitEdges()` and `ShapeUpgrade_WireDivide::Perform()` dereference `Context()` unconditionally. `ShapeFix_Root::Context()` returns `myContext`, which the base constructor leaves null and only an explicit, optional `SetContext()` ever fills — so calling `Init(...)` followed by `Perform()`, the usage the public API invites, is an immediate null-handle dereference: SIGSEGV at address 0, uncatchable in-process.

No unusual geometry is required. A plain 4-edge planar square face crashes both classes 100% of the time.

First unguarded dereference in each case (line numbers against current `master`):

| Entry point | Dereference |
|---|---|
| `ShapeUpgrade_WireDivide::Perform()` | `ShapeUpgrade_WireDivide.cxx:323` — `Context()->Apply(ItW.Value(), TopAbs_SHAPE)`, the first statement of the per-edge loop |
| `ShapeFix_ComposeShell::Perform()` | `ShapeFix_ComposeShell.cxx:506`, via `LoadWires()` — `Context()->Apply(iw.Value())` |
| `ShapeFix_ComposeShell::SplitEdges()` | the same `LoadWires()` dereference |

### Why this is not caught by the existing test suite

In-kernel, both classes are only ever driven by `ShapeUpgrade_FaceDivide::Perform()`, which opens with

```cpp
if (Context().IsNull())
{
}
```

and then hands its context down — to the compose shell at `ShapeUpgrade_FaceDivide.cxx:185`, to the wire divide at `:238`. Reached that way, neither class ever sees a null context. Reached directly through its own public `Init`/`Perform` pair, both crash on the first call.

Nine other classes in the same package already carry that self-creating guard: `ShapeFix_Shape::Init`, `ShapeFix_Shell::Perform`, `ShapeFix_Solid::Perform`, `ShapeFix_FixSmallFace::Init`, `ShapeFix_SplitCommonVertex::Init`, `ShapeFix_Wireframe` (both entry points), `ShapeFix_Wire::FixGap3d`/`FixGap2d`, `ShapeFix_Face::FixMissingSeam` and `ShapeUpgrade_ShapeDivide::Perform`. `ShapeFix_ComposeShell` and `ShapeUpgrade_WireDivide` are the two that were missed.

This is the same defect class as #1378, fixed for `ShapeFix_Face::FixPeriodicDegenerated` in #1380 (merged).

### Reproducer

```cpp
#include <BRepBuilderAPI_MakePolygon.hxx>
#include <BRep_Tool.hxx>
#include <Geom_Plane.hxx>
#include <NCollection_HArray2.hxx>
#include <ShapeExtend_CompositeSurface.hxx>
#include <ShapeFix_ComposeShell.hxx>
#include <ShapeUpgrade_WireDivide.hxx>
#include <TopExp_Explorer.hxx>
#include <TopoDS.hxx>
#include <gp_Pln.hxx>

int main()
{
  BRepBuilderAPI_MakePolygon p;
  p.Add(gp_Pnt(0, 0, 0));
  p.Add(gp_Pnt(10, 0, 0));
  p.Add(gp_Pnt(10, 10, 0));
  p.Add(gp_Pnt(0, 10, 0));
  p.Close();
  Handle(Geom_Plane) aPln = new Geom_Plane(gp_Pln(gp_Pnt(0, 0, 0), gp_Dir(0, 0, 1)));
  TopoDS_Face        aFace = BRepBuilderAPI_MakeFace(aPln, p.Wire(), Standard_True).Face();

  TopoDS_Wire aWire;
  for (TopExp_Explorer ex(aFace, TopAbs_WIRE); ex.More(); ex.Next())
  {
    aWire = TopoDS::Wire(ex.Current());
    break;
  }

  // (1) SIGSEGV: ShapeUpgrade_WireDivide.cxx:323, Context() is null
  Handle(ShapeUpgrade_WireDivide) aWD = new ShapeUpgrade_WireDivide();
  aWD->Init(aWire, aFace);
  aWD->Perform();

  // (2) SIGSEGV: ShapeFix_ComposeShell.cxx:506 via LoadWires(), Context() is null
  Handle(NCollection_HArray2<Handle(Geom_Surface)>) aGrid =
    new NCollection_HArray2<Handle(Geom_Surface)>(1, 1, 1, 1);
  aGrid->SetValue(1, 1, BRep_Tool::Surface(aFace));
  Handle(ShapeFix_ComposeShell) aCS = new ShapeFix_ComposeShell();
  aCS->Init(new ShapeExtend_CompositeSurface(aGrid), TopLoc_Location(), aFace, 1.0e-7);
  aCS->Perform();

  return 0;
}
```

Adding `SetContext(new ShapeBuild_ReShape)` before either `Perform()` call makes both complete normally, which is the workaround we currently carry downstream.

### Suggested fix

Add the same self-creating guard the nine sibling classes use, at the three public entry points missing it. `ShapeFix_ComposeShell`'s other context-dereferencing methods (`LoadWires`, `SplitWire`, `SplitByLine`, `MakeFacesOnPatch`, `DispatchWires`) are all reached through `Perform()` or `SplitEdges()`, so guarding those two covers them — and they are `const`, so they could not create a context themselves.

Verified downstream: with a context set (the only path that worked before), the resulting shape is byte-identical before and after the guard — `BRepTools::Write` dump hash plus face/wire/edge/vertex counts, for both a planar and an unbounded-cylindrical face — and the no-context path then produces that same result instead of crashing.

A PR follows.

### Environment

- OCCT `V8_0_0_p1`, reproduced against current `master` source as well (both entry points unchanged there)
- macOS 15 / arm64, clang, static build

gsdali:
Fixes #1409.

### Problem

`ShapeFix_ComposeShell::Perform()`, `ShapeFix_ComposeShell::SplitEdges()` and `ShapeUpgrade_WireDivide::Perform()` dereference `Context()` unconditionally. `ShapeFix_Root::Context()` returns `myContext`, which the base constructor leaves null and only an explicit, optional `SetContext()` ever fills — so `Init(...)` + `Perform()`, the usage both classes' public API invites, is a null-handle dereference (SIGSEGV at address 0, uncatchable in-process). A plain 4-edge planar square face reproduces it 100% of the time; see #1409 for a self-contained reproducer.

The reason this has gone unnoticed: in-kernel, both classes are only ever driven by `ShapeUpgrade_FaceDivide::Perform()`, which already self-creates a context and hands it down (`ShapeUpgrade_FaceDivide.cxx:185` for the compose shell, `:238` for the wire divide). Reached that way neither class ever sees a null context, so the existing tests never exercise the crash.

### Fix

Add the guard that nine other classes in the same package already use — `ShapeFix_Shape::Init`, `ShapeFix_Shell::Perform`, `ShapeFix_Solid::Perform`, `ShapeFix_FixSmallFace::Init`, `ShapeFix_SplitCommonVertex::Init`, `ShapeFix_Wireframe` (both entry points), `ShapeFix_Wire::FixGap3d`/`FixGap2d`, `ShapeFix_Face::FixMissingSeam`, `ShapeUpgrade_ShapeDivide::Perform` — to the three public entry points that were missing it:

```cpp
if (Context().IsNull())
{
}
```

`ShapeFix_ComposeShell`'s other context-dereferencing methods (`LoadWires`, `SplitWire`, `SplitByLine`, `MakeFacesOnPatch`, `DispatchWires`) are all reached through `Perform()` or `SplitEdges()`, so guarding those two covers them; they are also `const`, so they could not create a context themselves.

Same defect class as #1378 / #1380 (merged), which fixed `ShapeFix_Face::FixPeriodicDegenerated`.

### Validation

Verified downstream on `V8_0_0_p1` with the two patched translation units linked ahead of the stock archive, so only these symbols change:

- **Crash closed** — the reproducer from #1409, plus the same two calls against an unbounded-cylindrical face, SIGSEGV 100% of the time before and complete normally after.
- **No behaviour change on the working path** — with a context set (the only path that worked before), the result is byte-identical before and after: `BRepTools::Write` dump hash plus face/wire/edge/vertex counts, for both a planar and a cylindrical face, both classes. The guard only fires when there was no context, and in that case it produces exactly what an explicitly-provided context produces.
- `clang-format` clean on both files.

### DIFF
--- a/src/ModelingAlgorithms/TKShHealing/GTests/FILES.cmake
+++ b/src/ModelingAlgorithms/TKShHealing/GTests/FILES.cmake
@@ -7,8 +7,10 @@ set(OCCT_TKShHealing_GTests_FILES
   ShapeAnalysis_FreeBounds_Test.cxx
   ShapeBuild_ReShape_Test.cxx
   ShapeConstruct_ProjectCurveOnSurface_Test.cxx
+  ShapeFix_ComposeShell_Test.cxx
   ShapeFix_Face_Test.cxx
   ShapeFix_Shape_Test.cxx
   ShapeUpgrade_FaceDivide_Test.cxx
   ShapeUpgrade_UnifySameDomain_Test.cxx
+  ShapeUpgrade_WireDivide_Test.cxx
 )
--- a/src/ModelingAlgorithms/TKShHealing/GTests/ShapeFix_ComposeShell_Test.cxx
+++ b/src/ModelingAlgorithms/TKShHealing/GTests/ShapeFix_ComposeShell_Test.cxx
@@ -1,0 +1,108 @@
+// Copyright (c) 2026 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in OCCT
+// distribution for complete text of the license and disclaimer of any warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <gtest/gtest.h>
+
+#include <BRepBuilderAPI_MakeFace.hxx>
+#include <BRepBuilderAPI_MakePolygon.hxx>
+#include <BRep_Tool.hxx>
+#include <Geom_Plane.hxx>
+#include <Geom_Surface.hxx>
+#include <NCollection_HArray2.hxx>
+#include <ShapeBuild_ReShape.hxx>
+#include <ShapeExtend_CompositeSurface.hxx>
+#include <ShapeFix_ComposeShell.hxx>
+#include <TopLoc_Location.hxx>
+#include <TopoDS_Face.hxx>
+#include <gp_Pln.hxx>
+
+namespace
+{
+TopoDS_Face squareFace()
+{
+  BRepBuilderAPI_MakePolygon aPoly;
+  aPoly.Add(gp_Pnt(0, 0, 0));
+  aPoly.Add(gp_Pnt(10, 0, 0));
+  aPoly.Add(gp_Pnt(10, 10, 0));
+  aPoly.Add(gp_Pnt(0, 10, 0));
+  aPoly.Close();
+  occ::handle<Geom_Plane> aPln = new Geom_Plane(gp_Pln(gp_Pnt(0, 0, 0), gp_Dir(0, 0, 1)));
+  BRepBuilderAPI_MakeFace aMakeFace(aPln, aPoly.Wire(), true);
+  return aMakeFace.IsDone() ? aMakeFace.Face() : TopoDS_Face();
+}
+} // namespace
+
+// Test for issue #1409: ShapeFix_ComposeShell crashes with null context.
+TEST(ShapeFix_ComposeShellTest, Perform_WithoutContext_DoesNotCrash)
+{
+  const TopoDS_Face aFace = squareFace();
+  ASSERT_FALSE(aFace.IsNull());
+
+  occ::handle<Geom_Surface>                                   aSurf = BRep_Tool::Surface(aFace);
+  occ::handle<NCollection_HArray2<occ::handle<Geom_Surface>>> aGrid =
+    new NCollection_HArray2<occ::handle<Geom_Surface>>(1, 1, 1, 1);
+  aGrid->SetValue(1, 1, aSurf);
+  occ::handle<ShapeExtend_CompositeSurface> aCompSurf = new ShapeExtend_CompositeSurface(aGrid);
+
+  occ::handle<ShapeFix_ComposeShell> aShell = new ShapeFix_ComposeShell();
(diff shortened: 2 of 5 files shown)


dpasukhi:
Also please clean up code with too long comments

gsdali:
Thanks — pushed an update.

**GTests added**: `ShapeFix_ComposeShell_Test.cxx` and `ShapeUpgrade_WireDivide_Test.cxx` in `src/ModelingAlgorithms/TKShHealing/GTests/`, following the existing `ShapeUpgrade_FaceDivide_Test.cxx`'s pattern for this exact defect class (issue #1179). Each covers the entry point that was crashing (`Perform()` on both classes, plus `SplitEdges()` on `ShapeFix_ComposeShell`) with no context set, and a with/without-context equivalence check. Verified both directions before pushing: linked against a build with the guard reverted, all three `*_WithoutContext_DoesNotCrash` cases SIGSEGV (exit 139); linked against the guarded build, all five tests pass.

**Comments shortened**: dropped the three explanatory comments on the new guards entirely, matching the nine sibling guards already in this file/package (e.g. `ShapeFix_Shape::Init`), none of which carry one.

PR #1432, opened 2026-08-05
Modeling Algorithms - Fix ChFi2d_Builder::AddChamfer null-edge dereference on a repeated pair

### LINKED ISSUE #1431: ChFi2d_Builder::AddChamfer(edge, edge, ...) SIGSEGVs on a repeated edge pair (unchecked FindConnectedEdges status)
### Description

`ChFi2d_Builder::AddChamfer(const TopoDS_Edge& E1, const TopoDS_Edge& E2, double D1, double D2)`
calls `ChFi2d::FindConnectedEdges(newFace, commonVertex, EE1, EE2)` to look up the two edges
connected to the pair's shared vertex on the working face, then immediately dereferences the two
output edges (`EE1.IsSame(E2)`, and later passes both into `ComputeChamfer`) without checking the
returned status first:

```cpp
// ChFi2d_Builder_0.cxx, AddChamfer(E1, E2, D1, D2)
TopoDS_Edge EE1, EE2;
status = ChFi2d::FindConnectedEdges(newFace, commonVertex, EE1, EE2);
if (EE1.IsSame(E2))          // <-- no status check first
{
  ...
}
ComputeChamfer(commonVertex, EE1, EE2, D1, D2, E1Mod, E2Mod, chamfer);   // <-- null EE1/EE2 reach here
```

`ChFi2d::FindConnectedEdges` (`ChFi2d.cxx`) leaves both output edges null (default-constructed) on
every failure path, including the common one here: the vertex is simply not present in the current
face's vertex map, which returns `ChFi2d_ConnexionError` with neither edge assigned.

Calling `AddChamfer(E1, E2, D1, D2)` a second time with the *same* edge pair triggers exactly this:
the first call's `BuildNewWire` rebuilds the face's wire, replacing the pair's shared vertex, so the
second call's `FindConnectedEdges` on that (now stale) vertex fails, and the two null edges it
leaves behind are dereferenced.

The sibling overload, `AddChamfer(const TopoDS_Edge& E, const TopoDS_Vertex& V, double D, double Ang)`,
calls the identical `ChFi2d::FindConnectedEdges` and checks the status correctly:

```cpp
// ChFi2d_Builder_0.cxx, AddChamfer(E, V, D, Ang)
status = ChFi2d::FindConnectedEdges(newFace, V, adjEdge1, adjEdge2);
if (status == ChFi2d_ConnexionError)
{
  return aChamfer;
}
```

Reached from `BRepFilletAPI_MakeFillet2d::AddChamfer(E1, E2, D1, D2)`, and from OCCT's own DRAW
`chfi2d` command (`BRepTest_Fillet2DCommands.cxx`), which loops over edge-name pairs from the
command line and calls this same overload once per pair, so a `chfi2d` invocation naming the same
two edges twice in one call also reaches it.

### Expected Behavior

The second `AddChamfer(E1, E2, D1, D2)` call on the same pair should return a null edge and set
`Status()` to `ChFi2d_ConnexionError`, the same outcome `AddChamfer(E, V, D, Ang)` already gives for
an unconnected vertex.

### Actual Behavior

The process SIGSEGVs. Confirmed with a debug (`-O0`) single-TU override-link (the file compiled
standalone and linked before the OCCT static archive so the linker resolves this TU's symbols from
the override, not the stock library): the crash is inside `ComputeChamfer`, reached with `EE1`/`EE2`
both null.

```cpp
#include <BRepBuilderAPI_MakePolygon.hxx>
#include <BRepFilletAPI_MakeFillet2d.hxx>
#include <TopExp_Explorer.hxx>
#include <TopoDS.hxx>
#include <TopoDS_Edge.hxx>
#include <TopoDS_Face.hxx>
#include <gp_Pnt.hxx>
#include <vector>

int main() {
  BRepBuilderAPI_MakePolygon poly;
  poly.Add(gp_Pnt(0, 0, 0));
  poly.Add(gp_Pnt(10, 0, 0));
  poly.Add(gp_Pnt(10, 10, 0));
  poly.Add(gp_Pnt(0, 10, 0));
  poly.Close();
  TopoDS_Face face = BRepBuilderAPI_MakeFace(poly.Wire()).Face();

  std::vector<TopoDS_Edge> edges;
  for (TopExp_Explorer ex(face, TopAbs_EDGE); ex.More(); ex.Next())
    edges.push_back(TopoDS::Edge(ex.Current()));

  BRepFilletAPI_MakeFillet2d chamfer(face);
  chamfer.AddChamfer(edges[0], edges[1], 1.0, 2.0);   // first call: succeeds, Status() == ChFi2d_IsDone
  chamfer.AddChamfer(edges[0], edges[1], 1.0, 2.0);   // second call, SAME pair: *** SIGSEGV ***
  return 0;
}
```

Also reproducible via DRAW: `chfi2d result face e1 e2 CD 1 1 e1 e2 CD 1 1` (same two edge names
named twice) on any planar face with at least two connected edges.

### OCCT version

Reproduced on `V8_0_1` and confirmed present in the current `master` (the touched function is
byte-identical between the two).

### Platform

macOS arm64, but the defect is a null-pointer dereference reached through pure geometry, not
platform-specific code.

gsdali:
### Description of the change

`ChFi2d_Builder::AddChamfer(const TopoDS_Edge& E1, const TopoDS_Edge& E2, double D1, double D2)`
calls `ChFi2d::FindConnectedEdges` and dereferences the two edges it returns without checking the
returned status first. `FindConnectedEdges` leaves both edges null on every failure path, and a
second call naming the same edge pair hits exactly that: the pair's shared vertex is removed from
the face's wire by the first call's own `BuildNewWire`, so the second call's lookup fails and the
two null edges it returns reach `ComputeChamfer` unchecked.

Adds the same status check the sibling overload, `AddChamfer(const TopoDS_Edge& E, const
TopoDS_Vertex& V, double D, double Ang)`, already has immediately after the identical
`FindConnectedEdges` call:

```cpp
status = ChFi2d::FindConnectedEdges(newFace, commonVertex, EE1, EE2);
if (status == ChFi2d_ConnexionError)
{
  return chamfer;
}
```

`chamfer` is the default-constructed null edge this function already returns on its other refusal
paths (`ChFi2d_Builder_0.cxx` lines 83, 89, 95), so this reuses the existing "declined" signal
rather than introducing a new one.

### Related issue

Fixes #1431.

### Type of change

### How Has This Been Tested?

1. Standalone repro (attached in #1431): a rectangular planar face, `BRepFilletAPI_MakeFillet2d::AddChamfer`
   called twice with the identical edge pair. On stock `master`, the second call SIGSEGVs 100% of
   the time (`ComputeChamfer`, both edges null).
2. Debug (`-O0`) single-TU override-link (this file compiled standalone and linked before the OCCT
   static archive, so the linker resolves this TU's symbols from the override): before the patch,
   the second call crashes; after, it returns a null edge with `Status() == ChFi2d_ConnexionError`,
   matching the sibling overload's own answer for an unconnected vertex.
3. The first call in both cases is unaffected: `Status() == ChFi2d_IsDone`, non-null edge, same
   geometry before and after.
4. `clang-format --dry-run --Werror` on the touched file reports only pre-existing violations
   elsewhere in the file (`Geom2dInt_GInter` construction, unrelated lines), unchanged in count and
   content before and after this patch; the four added lines are clean.

### Checklist:

- [x] My code follows the code style of this project
- [x] I have commented my code, particularly in hard-to-understand areas — N/A, the added check
      mirrors the sibling overload's own existing, uncommented idiom line for line
- [x] My change requires a change to the documentation — N/A, no public API/behavior change beyond
      "does not crash"
- [x] I have added tests to cover my changes — regression coverage lives in the downstream Swift
      wrapper project (OCCTSwift), which found this crash through a fillet/chamfer contract audit;
      happy to add a DRAW test if maintainers point me at the right harness/fixture convention for
      this area

### DIFF
--- a/src/ModelingAlgorithms/TKFillet/ChFi2d/ChFi2d_Builder_0.cxx
+++ b/src/ModelingAlgorithms/TKFillet/ChFi2d/ChFi2d_Builder_0.cxx
@@ -99,6 +99,10 @@ TopoDS_Edge ChFi2d_Builder::AddChamfer(const TopoDS_Edge& E1,
   // on <newFace>
   TopoDS_Edge EE1, EE2;
   status = ChFi2d::FindConnectedEdges(newFace, commonVertex, EE1, EE2);
+  if (status == ChFi2d_ConnexionError)
+  {
+    return chamfer;
+  }
   if (EE1.IsSame(E2))
   {
     TopAbs_Orientation orient = EE1.Orientation();
--- a/src/ModelingAlgorithms/TKFillet/GTests/BRepFilletAPI_MakeFillet2d_Test.cxx
+++ b/src/ModelingAlgorithms/TKFillet/GTests/BRepFilletAPI_MakeFillet2d_Test.cxx
@@ -1,0 +1,67 @@
+// Copyright (c) 2026 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in OCCT
+// distribution for complete text of the license and disclaimer of any warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <BRepBuilderAPI_MakeFace.hxx>
+#include <BRepBuilderAPI_MakePolygon.hxx>
+#include <BRepFilletAPI_MakeFillet2d.hxx>
+#include <ChFi2d_ConstructionError.hxx>
+#include <TopExp_Explorer.hxx>
+#include <TopoDS.hxx>
+#include <TopoDS_Edge.hxx>
+#include <TopoDS_Face.hxx>
+#include <gp_Pnt.hxx>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+TopoDS_Face squareFace()
+{
+  BRepBuilderAPI_MakePolygon aPoly;
+  aPoly.Add(gp_Pnt(0, 0, 0));
+  aPoly.Add(gp_Pnt(10, 0, 0));
+  aPoly.Add(gp_Pnt(10, 10, 0));
+  aPoly.Add(gp_Pnt(0, 10, 0));
+  aPoly.Close();
+  BRepBuilderAPI_MakeFace aMakeFace(aPoly.Wire());
+  return aMakeFace.IsDone() ? aMakeFace.Face() : TopoDS_Face();
+}
+} // namespace
+
+// Regression test for issue #1431: a second AddChamfer() naming the same edge pair must not
+// crash. The pair's shared vertex is removed from the face's wire by the first call, so the
+// second call's FindConnectedEdges lookup fails; ComputeChamfer used to dereference the two
+// null edges it returned unchecked.
+TEST(BRepFilletAPI_MakeFillet2dTest, AddChamfer_RepeatedPair_DoesNotCrash)
+{
+  const TopoDS_Face aFace = squareFace();
+  ASSERT_FALSE(aFace.IsNull());
+
+  TopoDS_Edge     anEdge1, anEdge2;
+  TopExp_Explorer anExp(aFace, TopAbs_EDGE);
+  ASSERT_TRUE(anExp.More());
+  anEdge1 = TopoDS::Edge(anExp.Current());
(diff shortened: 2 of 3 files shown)


dpasukhi:

gsdali:
Thanks — pushed a GTest.

`BRepFilletAPI_MakeFillet2d_Test.cxx` in `src/ModelingAlgorithms/TKFillet/GTests/`, calling `AddChamfer` twice with the identical edge pair through the same public API the original crash was reported against. Verified both directions before pushing: linked with the guard removed, the second call SIGSEGVs (exit 139); linked with the fix, it returns a null edge with `Status() == ChFi2d_ConnexionError`, matching the sibling overload's existing answer for an unconnected vertex, and the test passes.

PR #1435, opened 2026-08-06
Modeling Data - Fix GeomTools_Curve2dSet/SurfaceSet null-handle Add()/Index() divergence from CurveSet

### LINKED ISSUE #1434: GeomTools_Curve2dSet::Add / GeomTools_SurfaceSet::Add accept a null handle and defer the crash to Write(), unlike the CurveSet sibling
### Description

`GeomTools_CurveSet::Add` guards against a null `Handle(Geom_Curve)` and drops it:

```cpp
// GeomTools_CurveSet.cxx
int GeomTools_CurveSet::Add(const occ::handle<Geom_Curve>& C)
{
  return (C.IsNull()) ? 0 : myMap.Add(C);
}
```

Its two siblings do not. `GeomTools_Curve2dSet::Add` and `GeomTools_SurfaceSet::Add` accept a null
handle silently and hand back a valid-looking index:

```cpp
// GeomTools_Curve2dSet.cxx
int GeomTools_Curve2dSet::Add(const occ::handle<Geom2d_Curve>& S)
{
  return myMap.Add(S);
}

// GeomTools_SurfaceSet.cxx
int GeomTools_SurfaceSet::Add(const occ::handle<Geom_Surface>& S)
{
  return myMap.Add(S);
}
```

The caller gets no signal at that point at all. The crash comes later, inside `Write()`:
`Curve2dSet::Write` → `PrintCurve2d` → `C->DynamicType()`, and `SurfaceSet::Write` → `PrintSurface`
→ `S->DynamicType()`, both dereferencing the null handle that `Add()` accepted.

The same asymmetry repeats one function down: `Index()`. `CurveSet::Index` also guards; the two
siblings do not:

```cpp
int GeomTools_CurveSet::Index(const occ::handle<Geom_Curve>& S) const
{
  return S.IsNull() ? 0 : myMap.FindIndex(S);
}

int GeomTools_Curve2dSet::Index(const occ::handle<Geom2d_Curve>& S) const
{
  return myMap.FindIndex(S);
}

int GeomTools_SurfaceSet::Index(const occ::handle<Geom_Surface>& S) const
{
  return myMap.FindIndex(S);
}
```

`Index()` does not crash (`NCollection_IndexedMap::FindIndex` never dereferences its argument, it
only hashes/compares the handle's identity), but it silently reports a non-zero index for a handle
that was never validly bound, where `CurveSet::Index` correctly reports 0.

### Expected Behavior

`GeomTools_Curve2dSet::Add`/`GeomTools_SurfaceSet::Add` should drop a null handle and return 0,
matching `GeomTools_CurveSet::Add`. `Index()` should return 0 for a null handle on all three,
matching `GeomTools_CurveSet::Index`.

### Actual Behavior

`Add()` accepts the null and returns a bound index (`1` on a fresh set); `Write()` then SIGSEGVs
dereferencing it. `Index()` returns the same bogus non-zero index instead of 0.

```cpp
#include <GeomTools_Curve2dSet.hxx>
#include <GeomTools_CurveSet.hxx>
#include <GeomTools_SurfaceSet.hxx>
#include <Geom_Curve.hxx>
#include <Geom_Surface.hxx>
#include <Geom2d_Curve.hxx>
#include <sstream>

int main() {
  occ::handle<Geom_Curve>   nullCurve3d;
  occ::handle<Geom2d_Curve> nullCurve2d;
  occ::handle<Geom_Surface> nullSurface;

  GeomTools_CurveSet cs;
  printf("CurveSet::Add(null) = %d\n", cs.Add(nullCurve3d));      // 0, dropped
  std::ostringstream o1; cs.Write(o1);                             // fine

  GeomTools_Curve2dSet c2s;
  printf("Curve2dSet::Add(null) = %d\n", c2s.Add(nullCurve2d));   // 1, accepted
  std::ostringstream o2; c2s.Write(o2);                            // *** SIGSEGV ***

  GeomTools_SurfaceSet ss;
  printf("SurfaceSet::Add(null) = %d\n", ss.Add(nullSurface));    // 1, accepted
  std::ostringstream o3; ss.Write(o3);                             // *** SIGSEGV ***
  return 0;
}
```

Measured output on the affected kernel:

```
  -- Add() alone --
  GeomTools_CurveSet::Add(null)                  Add returned 0    returned normally
  GeomTools_Curve2dSet::Add(null)                Add returned 1    returned normally
  GeomTools_SurfaceSet::Add(null)                Add returned 1    returned normally

  -- Add() then Write(), which is what the bridge does --
  GeomTools_CurveSet::Add(null) + Write          returned normally
  GeomTools_Curve2dSet::Add(null) + Write        SIGSEGV (uncatchable)
  GeomTools_SurfaceSet::Add(null) + Write        SIGSEGV (uncatchable)
```

### OCCT version

Reproduced on `V8_0_1` and confirmed present in current `master` (the three touched files are
byte-identical between the two).

### Platform

macOS arm64, but the defect is a null-pointer dereference reached through pure container logic, not
platform-specific code.

gsdali:
### Description of the change

`GeomTools_CurveSet::Add` guards a null `Handle(Geom_Curve)` and drops it (`return (C.IsNull()) ? 0
: myMap.Add(C);`). `GeomTools_Curve2dSet::Add` and `GeomTools_SurfaceSet::Add` do not: they accept
the null and hand back a valid-looking index, which later crashes `Write()` when it dereferences the
handle it was never checked against. `Index()` has the identical divergence one function down
(`CurveSet::Index` guards, the other two do not) — it does not crash, since
`NCollection_IndexedMap::FindIndex` only hashes/compares the handle's identity, but it silently
returns a bogus non-zero index for a handle that was never validly bound.

Adds the same guard `CurveSet::Add`/`CurveSet::Index` already have, to both methods on both sibling
classes — four one-line changes, matching the correct sibling rather than introducing a new pattern:

```cpp
// GeomTools_Curve2dSet.cxx
int GeomTools_Curve2dSet::Add(const occ::handle<Geom2d_Curve>& S)
{
  return (S.IsNull()) ? 0 : myMap.Add(S);
}
...
int GeomTools_Curve2dSet::Index(const occ::handle<Geom2d_Curve>& S) const
{
  return (S.IsNull()) ? 0 : myMap.FindIndex(S);
}

// GeomTools_SurfaceSet.cxx
int GeomTools_SurfaceSet::Add(const occ::handle<Geom_Surface>& S)
{
  return (S.IsNull()) ? 0 : myMap.Add(S);
}
...
int GeomTools_SurfaceSet::Index(const occ::handle<Geom_Surface>& S) const
{
  return (S.IsNull()) ? 0 : myMap.FindIndex(S);
}
```

### Related issue

Fixes #1434.

### Type of change

### How Has This Been Tested?

1. Standalone repro (attached in #1434): `Add(nullHandle)` then `Write()` on `GeomTools_Curve2dSet`
   and `GeomTools_SurfaceSet` SIGSEGVs on stock `master`/`V8_0_1`; `GeomTools_CurveSet` does not,
   since its `Add()` already drops the null.
2. Override-link (this repo's two changed files compiled standalone and linked ahead of the OCCT
   static archive, so the linker resolves these two TUs' symbols from the override): before the
   patch, `Curve2dSet`/`SurfaceSet` both accept the null (`Add` returns 1) and SIGSEGV in `Write()`;
   after, all three classes' `Add()` return 0 and `Write()` completes normally on all three.
3. `Index()` before the patch returns the same bogus 1 for `Curve2dSet`/`SurfaceSet` after adding a
   null handle; after, all three classes' `Index()` return 0, matching `CurveSet::Index`.
4. A populated, all-valid-handle set is unaffected in either direction: `Add`/`Index`/`Write`/`Read`
   behavior for non-null handles is untouched by this change (the new branch is only taken when
   `IsNull()` is true).

### Checklist:

- [x] My code follows the code style of this project
- [x] I have commented my code, particularly in hard-to-understand areas — N/A, both added checks
      mirror `GeomTools_CurveSet`'s own existing, uncommented idiom line for line
- [x] My change requires a change to the documentation — N/A, no public API/behavior change beyond
      "does not crash, and reports the same index CurveSet already reports for a null handle"
- [x] I have added tests to cover my changes — regression coverage lives in the downstream Swift
      wrapper project (OCCTSwift), which found this crash through a null-handle-guard census; happy
      to add a GTest if maintainers point me at the right fixture convention for this area


dpasukhi:

gsdali:
Thanks — pushed GTests.

`GeomTools_Curve2dSet_Test.cxx` and `GeomTools_SurfaceSet_Test.cxx` in `src/ModelingData/TKGeomBase/GTests/`, four cases each: `Add()`/`Index()` with a null handle (must drop it / return 0, matching `GeomTools_CurveSet`), `Write()` after adding a null (must not crash), and a valid-handle case as a control. Verified both directions: against the current pinned kernel (predates this fix), `Add(null)` returns 1 and `Write()` SIGSEGVs; linked with the patched translation units ahead of the archive, all 8 tests pass.

PR #1449, opened 2026-08-08
Modeling Algorithms - Enhance fillet and chamfer handling. Part 2

### LINKED ISSUE #172: Fillet can not round a face
### Category

Modeling

### Severity

Major

### Description

I know this bug is in the original tracker [25478](https://tracker.dev.opencascade.org/view.php?id=25478)
but it has been 10 years since it was reported and there is no sign of any attention or progress done on this. Please someone needs to look into it since it is feels like the most important problem now in FreeCAD

### Expected Behavior

if the fillet radius is big enough that there are no linear faces remaining, it should round the face.

### Actual Behavior

The fillet will work as long as the radius is small enough to keep some face remaining. if it is big it will not work.

see original issue for more description and examples. [25478](https://tracker.dev.opencascade.org/view.php?id=25478)

### Operating System

Linux

### Compiler

GCC

### Compiler Version

_No response_

### Bitness

64-bit

### OCCT Version

latest

### Precise OCCT Version or Branch

_No response_

### Environment

_No response_

### Additional Files

_No response_

dpasukhi:
- Detect complete coincidence between fillet and chamfer intersection curves.
- Preserve orientation relationships between equivalent intersection curves.
- Reuse existing restriction edges instead of constructing duplicate topology.
- Merge coincident curve endpoints into shared topological vertices.
- Exclude fully consumed faces and their obsolete boundary edges during reconstruction.
- Mark edges collapsed to a single vertex as degenerated during construction.
- Accept tangential contacts while rejecting transversal intersections.
- Preserve generated and modified shape history in the resulting topology.
- Add GTests for convex, concave, opposing-edge, chamfer, and history cases.

### DIFF
--- a/src/ModelingAlgorithms/TKBool/GTests/FILES.cmake
+++ b/src/ModelingAlgorithms/TKBool/GTests/FILES.cmake
@@ -6,4 +6,5 @@ set(OCCT_TKBool_GTests_FILES
   BRepAlgoAPI_Fuse_Test.cxx
   BRepAlgoAPI_Section_Test.cxx
   BRepFill_PipeShell_Test.cxx
+  TopOpeBRepDS_BuildTool_Test.cxx
 )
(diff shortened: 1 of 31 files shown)


dpasukhi:
image
image
image
image
image

PR #1450, opened 2026-08-09
Mesh - Fix iterator invalidation in Delaun triangle-kill cascade

bwross:
killTrianglesAroundVertex() and killTrianglesOnIntersectingLinks() each iterated myMeshData->LinksConnectedTo(node) with a live iterator while their own loop bodies (via killTrianglesOnIntersectingLinks() and killLinkTriangles()) can delete triangles and edges connected to that same node, mutating the list a live iterator would be walking. Continuing to use such an iterator after the list is mutated out from under it is undefined behavior.

Fixed by snapshotting each node's neighbor link ids into a stable vector before iterating, skipping any that turn out already-deleted by the time they're reached. This is safe because nothing in this whole kill-cascade ever calls AddLink(), so no link id snapshotted here can be recycled for an unrelated edge before it's reached.

Verified byte-identical mesh output (231 faces, 33312 triangles, 32874 nodes, same max deflection) on a 225-hole plate stress shape chosen to drive many vertex insertions through this kill-cascade, both before and after the fix. Full mesh DRAW test suite passes (SKIPPED cases are missing external test-data files, a documented non-regression status in OCCT's own test methodology).

CLA ID is 1145.

### DIFF
--- a/src/ModelingAlgorithms/TKMesh/BRepMesh/BRepMesh_Delaun.cxx
+++ b/src/ModelingAlgorithms/TKMesh/BRepMesh/BRepMesh_Delaun.cxx
@@ -1543,18 +1543,42 @@ void BRepMesh_Delaun::killTrianglesAroundVertex(
   IMeshData::MapOfIntegerInteger&     theLoopEdges,
   IMeshData::VectorOfInteger&         theVictimNodes)
 {
-  IMeshData::ListOfInteger::Iterator aNeighborsIt = myMeshData->LinksConnectedTo(theZombieNodeId);
+  // Snapshot the neighbor link ids before iterating: the loop body below can,
+  // via killTrianglesOnIntersectingLinks()/killLinkTriangles(), delete
+  // triangles and edges connected to theZombieNodeId, which mutates
+  // myMeshData->LinksConnectedTo(theZombieNodeId), the very list a live
+  // iterator would be walking. Continuing to use such an iterator after the
+  // list is mutated out from under it is undefined behavior. Iterating a
+  // stable copy instead, and skipping entries that have already become
+  // BRepMesh_Deleted by the time we reach them, is safe here specifically
+  // because nothing in this whole kill-cascade calls AddLink(), so no link
+  // id snapshotted here can be recycled for an unrelated edge before we get
+  // to it.
+  IMeshData::VectorOfInteger aNeighborLinkIds;
+  {
+    IMeshData::ListOfInteger::Iterator aLinksIt = myMeshData->LinksConnectedTo(theZombieNodeId);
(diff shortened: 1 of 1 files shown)


dpasukhi:
Thank you for your patch, during today I'm planning to a little refactor code without critical changes.
Planned to be merged to IR during today/tomorrow.

Many thanks for your contribution!

dpasukhi (approved):

dpasukhi:
Thank you for the patch

Mr-Question:
@dpasukhi, @bwross,

Hi guys, I think that by that patch you have actually got rid of a small, but still the optimization.

The problem described in the PR (undefined behavior) should never happen at all, because of the fact that _BRepMesh_DataStructureOfDelaun_, the one that keeps info about the nodes, is initialized using common memory inc allocator (if you would look outside the _BRepMesh_Delaun_, namely to _BRepMesh_BaseMeshAlgo_ to have a bigger picture). That thing never (or never before...) released memory until it is requested explicitly or on its destruction.

Like this, even removed from the list, content of a list node, i.e. pointer to the next element, remains valid in memory and can be used to continue iteration without fear that it has become invalid as all next pointers also remain valid by definition.

Previous pointers could not cause any harm as they should not appear in the iteration on the next call to _linksConnectedTo()_ because they are already removed; next list nodes, if they are removed, should not appear in the iteration as well, because pointers to the next element in the list should be properly updated which means current iterator could not be invalidated if it stays before the removed elements.

Current patch causes allocation of an additional memory on the heap for every instance of _NCollection_Array1_ and adds iteration to copy content of the list which could cause performance slow down on the large models, and a check. 

That is very subtle optimization, so the patch is not meaningless in general as it removes that implicit specific and makes code clear for the reader, but it also influence performance as the optimization was obviously intended to avoid exactly that same problem taking into account the advanced details of the implementation.

I understand that to push the solution forward, OCCT does not require any evidence to prove the existence of an abstract problem now, but yet it would be interesting to see the real use case that caused this change to happen.

More than that, it would be useful for OCCT itself as you can include proper test case to non-regression data base (if there is any).

If there was a real use case I wonder why the PR would have gone without official public issue describing the problem with data to reproduce the reported issue. If there was not, there should be nothing to worry about and it would be better just to add a comment about this opt.

My kind regards.

dpasukhi:
Dear  @Mr-Question  thank you a lot for your analysing!

I totally agree about the "surviving of the data". List uses IncAllocator and indeed iterators always valid. Many thanks!
I will add some extra tests to keep behaviour with guarantee surviving of nodes in case of keeping memory alive. (On List contaienrs itself, generic one)

Many thanks again! I always happy to have constructive feedback ;)
In case if you will have anything similar, I would be happy to hear it, but it is not necessary to describe so detailed :)

PR #1462, opened 2026-08-14
Modeling Algorithms - Guard silently-uncapped solid in BRepOffsetAPI_ThruSections

gsdali:
## Pre-Submission Checks

## Problem / Motivation

`BRepOffsetAPI_ThruSections(isSolid: true)` silently omits both end-cap faces when a closed
section wire has two or more periods of out-of-plane variation around the loop, a genuinely
non-planar closed curve. `Build()` still calls `Done()` and reports `IsDone() == true`, and
`Shape()` returns a `TopoDS_Solid` that is missing two faces and is not actually closed.
`BRepCheck_Analyzer` catches this later, but with no localized error (`errorCount == 0`, no
detailed check statuses), so a caller has already trusted `IsDone()` by the time anything flags
it.

Root cause: `MakeSolid()` (the static helper both `CreateRuled()` and `CreateSmoothed()` call to
close a loft's two open ends) caps each end via `PerformPlan()`, which tries
`BRepBuilderAPI_FindPlane` and then a `BRepLib_FindSurface`-backed `MakeFace(wire)` fallback.
Neither finds anything but a plane or a surface already attached to the wire's edges. A section
wire with `k >= 2` periods of out-of-plane variation around the loop has no such surface, so
`PerformPlan()` fails for it. `MakeSolid()` already tracks this in its own local `B`, threaded
through both `PerformPlan()` calls — and discards it: it force-marks the shell and solid
`Closed(true)` unconditionally before returning, and the caller then calls `Done()`
unconditionally too.

A single period of out-of-plane variation (e.g. `z = amp * cos(theta)` at constant radius) caps
fine, because that curve is secretly planar: it is exactly the intersection of the cylinder
`r = const` with a tilted plane, so `BRepBuilderAPI_FindPlane` finds it. `k >= 2` has no such
plane and no other analytic surface either.

Found downstream of a bevel gear `ThruSections` loft in the OCCTSwift ecosystem, where each
section is a toothed profile whose out-of-plane period equals the tooth count, always `>= 2` for
a real gear.

## Proposed Solution

`MakeSolid()`'s local `B` already distinguishes exactly the two cases that matter: it stays
`true` when capping either succeeded or wasn't needed (`PerformPlan()`'s own degenerate-wire
shortcut — every edge of the wire is degenerate, a single-vertex "point" section built via
`AddVertex()`, e.g. a cone's apex, which needs no cap face and correctly reports success with a
null face), and ends `false` only when capping was genuinely attempted and failed. After the
capping block, if `B` is still `false`, throw `StdFail_NotDone`, matching the exception this same
function already throws for a null shell two lines above:

```cpp
if (!B)
{
  throw StdFail_NotDone("BRepOffsetAPI_ThruSections: could not close a non-planar extremity");
}
```

No signature change and no call-site change: both call sites already run inside `Build()`'s only
`try`/`catch` (`catch (Standard_Failure const&) { NotDone(); return; }`), so the throw propagates
there and `IsDone()` correctly reads `false`.

This does not make `BRepOffsetAPI_ThruSections` able to cap a genuinely non-planar wire (that
would need a real capping algorithm, e.g. an N-sided patch, out of scope here). It converts a
silent wrong success into a correctly reported failure, matching every other failure path in this
file.

A null-face check (`myFirst.IsNull() || myLast.IsNull()`) looks like an equally natural fix and
is not: `PerformPlan()` also leaves the output face null, and returns `true` (no failure), when
the wire is degenerate, so a null-face check cannot tell "no cap needed" apart from "cap needed
and failed." Checked directly against `BOPAlgo_PaveFillerTest.FuseConeLoftWithBox_DegeneratedEdge`
(a circle-to-vertex loft, in this same PR's GTest additions) before settling on the `B`-based fix
— a null-face guard regresses that test, `B` does not.

**Limitation:** `GetStatus()` is not updated by this fix and stays
`BRepFill_ThruSectionErrorStatus_Done` on this path. `Build()`'s catch block doesn't touch
`myStatus` for any exception source, not one this fix introduces — `IsDone()` is the correct,
fixed signal.

## Validation

Checks performed:

- [x] `clang-format --dry-run --Werror` against this tree's own `.clang-format`: zero violations
      on both changed files.
- [x] Added GTest cases in `BRepOffsetAPI_ThruSections_Test.cxx`:
      `NonPlanarClosedWireCappingFails` (a two-wire `k=2` loft, must now fail) and
      `DegenerateVertexEndStillSucceeds` (a circle-to-vertex loft, must still succeed — the same
      shape as `BOPAlgo_PaveFillerTest.FuseConeLoftWithBox_DegeneratedEdge`, added as a direct,
      local regression guard for it in this file too).
- [x] CI on this branch (macOS/Linux/Windows builds, all `Run GTest`/`Test`/`Regression Test`
      jobs including the No-PCH debug build) is green.
- [x] Locally, before pushing: override-link tested (compile the changed `.cxx` standalone, link
      it ahead of a prebuilt `libOCCT-macos.a`, no full OCCT rebuild) both an unpatched and a
      patched build. Against the unpatched build, `NonPlanarClosedWireCappingFails` fails exactly
      as described above (`IsDone()` reads `true`). Against the patched build, it passes, along
      with `DegenerateVertexEndStillSucceeds` and all pre-existing tests in the file. Also
      compiled and ran the actual `BOPAlgo_PaveFiller_Test.cxx`
      `FuseConeLoftWithBox_DegeneratedEdge` test, unmodified, against the patched build: passes.


gsdali:
Pushed a small strengthening found during downstream review (SecondMouseAU/OCCTSwift#909): if
wire1's `PerformPlan()` succeeds but wire2's fails, `face1` (aliased to `myFirst`, exposed via
`FirstShape()`) could still hold a real face that was never added to the shell, even though the
overall build correctly reports `IsDone() == false`. Nullifying `face1`/`face2` in the failure
branch closes that. No change to the core logic (`B`, the throw, the exception type) — re-verified
with the same override-link technique against both the two new GTests and
`FuseConeLoftWithBox_DegeneratedEdge`, all still pass.

PR #1486, opened 2026-08-22
Modeling - Fix Intersection join shell on BSpline loft faces

angelobartolome:
## Pre-Submission Checks

## Problem / Motivation

`BRepOffsetAPI_MakeThickSolid` with `GeomAbs_Intersection` can silently keep the opening face when shelling a solid whose wall is a bent BSpline (NURBS loft), even though a slightly smaller offset on the same shape succeeds.

On a 90° NURBS loft of two disks (r=30 in XY, r=15 in XZ), offset **-0.5** opens the top face and yields a valid 5-face shell. Offset **-0.8** still reports `IsDone()` / `NoError`, but glue sees the same face count as the input solid, zeroes the offset, and returns the original solid with the opening still there. The sampled min radius of curvature is about 6.6 mm, so -0.8 is not past a curvature limit.

User-visible impact: Intersection-join shells of NURBS lofts drop offset walls and look like a no-op instead of an open shell.

## Proposed Solution

Two targeted changes in the Intersection-join path:

1. **Enlarge closing faces for both offset signs.** `BuildOffsetByInter` used to enlarge context/opening faces only when `myOffset > 0`. A bent wall can push the wall–opening intersection outside the original disk for inward offsets as well. Enlargement now runs whenever closing faces are present (`!myFaces.IsEmpty()`).

2. **Approximate BSpline/Bezier `Geom_OffsetSurface` to a BSpline** when `JoinType == GeomAbs_Intersection`. IntWalk / 2D pcurves / `MakeLoops` on a raw offset NURBS often fail to close (one of two enlarged faces gets no image). The approximated surface keeps the same UV domain so existing pcurves remain valid. Plane/cone offsets are unchanged; Arc join still uses `Geom_OffsetSurface`. Conversion reuses `ShapeConstruct::ConvertSurfaceToBSpline` with the same C0 / `Precision::Approximation()` / 10000 / 15 settings `ShapeCustom_ConvertToBSpline` already uses for offset faces (C1 can hang in `GeomConvert_ApproxSurface`). On a null result, the original offset surface is kept.

After these changes, -0.8 Intersection on the loft rebuilds **3 offset faces**, glues to a **5-face** thick solid, removes the opening, and passes `BRepCheck`.

image

## Validation

Checks performed:

- `BRepOffsetAPI_MakeThickSolidTest.LoftTwoCircles_ShellTopFace` — 90° loft (disk r=30 in XY, r=15 in XZ), **-0.8 Intersection**: done, valid, opening removed

New GTest: `src/ModelingAlgorithms/TKOffset/GTests/BRepOffsetAPI_MakeThickSolid_Test.cxx`. `LoftTwoCircles_ShellTopFace` builds a 90° NURBS loft from two disks and asserts -0.8 Intersection opens the top face.

## CLA Confirmation

- [x] I confirm that I have read the contribution requirements, and that I have signed and submitted the CLA or I am covered by an approved company CLA.

CLA ID / submission reference: 1144

Pending

## Review Notes

- **Scope of approximation:** only Intersection join, and only when the offset basis is BSpline or Bezier with finite bounds. Analytic offsets (plane, cylinder, cone, …) are untouched. Arc join is unchanged and can still produce a BRepCheck-invalid result on this loft.
- **Reuse:** approximation is `ShapeConstruct::ConvertSurfaceToBSpline` (already used by `ShapeCustom_ConvertToBSpline` for offset faces). TKOffset already depends on TKShHealing. Face tolerance is raised to at least `Precision::Approximation()` because that helper does not return `MaxError`.
- **Performance:** conversion adds cost on Intersection join of NURBS faces. It is not applied to Arc join (the common default).
- **Compatibility:** no API change. Result geometry of Intersection-join NURBS offsets may differ slightly from a true `Geom_OffsetSurface` within `Precision::Approximation()`.
- **Tests:** `LoftTwoCircles_ShellTopFace` constructs the loft (disk r=30 in XY at `(0, -65, 0)`, disk r=15 in XZ at `(0, 0, 70)`) and shells the highest-Z face.

### DIFF
--- a/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_MakeOffset.cxx
+++ b/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_MakeOffset.cxx
@@ -1351,11 +1351,7 @@ void BRepOffset_MakeOffset::BuildOffsetByInter(const Message_ProgressRange& theR
   //-------------------------------------------------------------------
   // Extension of faces and calculation of new edges of intersection.
   //-------------------------------------------------------------------
-  bool ExtentContext = false;
-  if (myOffset > 0)
-  {
-    ExtentContext = true;
-  }
+  const bool ExtentContext = !myFaces.IsEmpty();
 
   BRepOffset_Inter3d Inter3(AsDes, Side, myTol);
   // Intersection between parallel faces
--- a/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_MakeOffset.cxx
+++ b/src/ModelingAlgorithms/TKOffset/BRepOffset/BRepOffset_MakeOffset.cxx
@@ -60,10 +60,13 @@
 #include <Geom2d_Line.hxx>
 #include <Geom2d_TrimmedCurve.hxx>
 #include <Geom2dAdaptor_Curve.hxx>
+#include <Geom_BezierSurface.hxx>
+#include <Geom_BSplineSurface.hxx>
 #include <Geom_Circle.hxx>
 #include <Geom_ConicalSurface.hxx>
 #include <Geom_OffsetSurface.hxx>
 #include <Geom_Plane.hxx>
+#include <Geom_RectangularTrimmedSurface.hxx>
 #include <Geom_SphericalSurface.hxx>
 #include <Geom_TrimmedCurve.hxx>
 #include <GeomAdaptor_Surface.hxx>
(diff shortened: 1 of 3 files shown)


angelobartolome:
Looks like this is too agressive and broke other offsets 🤔

dpasukhi:
> Looks like this is too agressive and broke other offsets 🤔

Yes, converting to BSpline can have some consequences because BSpline is complex type to work with, I suppose :(

angelobartolome:
Closing this to revisit later with a better approach, will open an issue to track it better

PR #1492, opened 2026-08-23
Modeling Algorithms - Poll the progress range for a user break in ChFi3d_Builder::Compute

Artem-Viveritsa:
## Pre-Submission Checks

## Problem / Motivation

`BRepFilletAPI_MakeFillet::Build` and `BRepFilletAPI_MakeChamfer::Build` accept a `Message_ProgressRange` and drop it. The argument name is commented out in the signature itself, because `ChFi3d_Builder::Compute` has nowhere to take it:

```cpp
void BRepFilletAPI_MakeFillet::Build(const Message_ProgressRange& /*theRange*/)
{
  myBuilder.Compute();   // the range does not go any further
  ...
}
```

There is no `UserBreak` anywhere in `ChFi3d`. Booleans, meshing, offsetting and shape healing all poll for a user break; fillet and chamfer are the ones that cannot be interrupted once started. An application that runs a fillet on a background thread has no way to give up on it — the only remaining options are to wait, or to kill the process.

That matters because a fillet build can run for a long time on input that is not obviously pathological, and an interactive caller (a radius dragged in a UI, a preview recomputed per frame) needs to abandon superseded work rather than accumulate it.

## Proposed Solution

`ChFi3d_Builder::Compute` now takes a `Message_ProgressRange`, defaulted to `Message_ProgressRange()` so every existing caller keeps compiling and behaving exactly as before.

It polls `UserBreak()` at the three places the algorithm is already between units of work, so no partially built unit is ever abandoned mid-way:

1. before each contour, at the head of the `myListStripe` loop, ahead of `PerformSetOfSurf`;
2. before each vertex, at the head of the `myVDataMap` loop, ahead of `PerformFilletOnVertex`;
3. before the topological reconstruction.

On a break the builder sets `done = false` and returns, so `IsDone()` reports false and the caller gets no half-made shape. `Reset()` puts the maker back to work as usual.

Both `Build` methods now pass their range down. Progress *reporting* is deliberately not added here: the fillet has no cheap way to state how much work is left, and a break poll is the part callers actually need. The parameter is in place should someone want to scale the range later.

## Validation

Three GTests added, covering both makers:

- `BRepFilletAPI_MakeFilletTest.UserBreakGivesUpTheBuild` — a break asked for before the first contour leaves the maker not done after exactly one poll, and `Reset()` lets the same maker build successfully again.
- `BRepFilletAPI_MakeFilletTest.UnbrokenProgressRangeBuildsAsBefore` — a range that never breaks produces the same valid result as before, and all three poll points are reached.
- `BRepFilletAPI_MakeChamferTest.UserBreakGivesUpTheBuild` — the same for the chamfer builder, which shares `Compute`.

Local build: minimal-module `Release`, `FoundationClasses` + `ModelingData` + `ModelingAlgorithms`.

- New GTests: 3/3 pass.
- Fillet-related suites (`*Fillet*`, `*Chamfer*`, `*ChFi*`, `*Blend*`, `*Offset*`): 248/249 pass, 1 pre-existing skip (`BRepOffsetAPI_MakePipeShellTest.Bug24909_2_ClosedCircularSpine`).
- Full `OpenCascadeGTest`: **7574/7579 pass, 0 failures**; the 5 skips are pre-existing and unrelated.

> Both figures were measured with this change and its companion PR applied together; the two are independent and touch disjoint code, but I did not run the suite with only one of them in the tree.

Checks performed:

- [x] `clang-format --style=file` applied to all changed files.

## CLA Confirmation

- [ ] I confirm that I have read the contribution requirements, and that I have signed and submitted the CLA or I am covered by an approved company CLA.

CLA ID / submission reference: **not signed yet — please tell me where to send it.**

I have read `.github/CLA.md` and am ready to sign, but every route `CONTRIBUTING.md` gives for submitting it currently 301-redirects to `https://occt3d.com/open-cascade-technology/index.html`, a page with no contribution or CLA section at all:

- the submission form, `https://dev.opencascade.org/get_involed/cla_submission_form`
- the portal root, `https://dev.opencascade.org`
- the agreement PDF linked from `.github/CLA.md` itself, `https://dev.opencascade.org/sites/default/files/Contributor_License_Agreement.pdf`

Point me at a current form or address and I will sign and submit before you merge; happy to have this PR held until then. Either way the stale links are probably worth fixing in `CONTRIBUTING.md` and `CLA.md`.

## Review Notes

- The default argument keeps this source- and behaviour-compatible for every existing caller of `Compute()`.
- Touches `ChFi3d_Builder.cxx` near the `ChFi3d_InitChron(cl_filds)` block, which #1490 also edits. No semantic conflict; whichever lands second may need a two-line context rebase.
- Companion PR #1491 bounds an unbounded probe loop in `ChFi3d_Builder_6.cxx::StoreData` — that fixes one specific stall, while this PR gives callers a way out of any other.

### DIFF
--- a/src/ModelingAlgorithms/TKFillet/BRepFilletAPI/BRepFilletAPI_MakeChamfer.cxx
+++ b/src/ModelingAlgorithms/TKFillet/BRepFilletAPI/BRepFilletAPI_MakeChamfer.cxx
@@ -272,9 +272,9 @@ occ::handle<TopOpeBRepBuild_HBuilder> BRepFilletAPI_MakeChamfer::Builder() const
 
 //=================================================================================================
 
-void BRepFilletAPI_MakeChamfer::Build(const Message_ProgressRange& /*theRange*/)
+void BRepFilletAPI_MakeChamfer::Build(const Message_ProgressRange& theRange)
 {
-  myBuilder.Compute();
+  myBuilder.Compute(theRange);
   if (myBuilder.IsDone())
   {
     Done();
--- a/src/ModelingAlgorithms/TKFillet/BRepFilletAPI/BRepFilletAPI_MakeFillet.cxx
+++ b/src/ModelingAlgorithms/TKFillet/BRepFilletAPI/BRepFilletAPI_MakeFillet.cxx
@@ -368,9 +368,9 @@ occ::handle<TopOpeBRepBuild_HBuilder> BRepFilletAPI_MakeFillet::Builder() const
 
 //=================================================================================================
 
-void BRepFilletAPI_MakeFillet::Build(const Message_ProgressRange& /*theRange*/)
+void BRepFilletAPI_MakeFillet::Build(const Message_ProgressRange& theRange)
 {
-  myBuilder.Compute();
+  myBuilder.Compute(theRange);
   if (myBuilder.IsDone())
   {
     Done();
--- a/src/ModelingAlgorithms/TKFillet/ChFi3d/ChFi3d_Builder.cxx
+++ b/src/ModelingAlgorithms/TKFillet/ChFi3d/ChFi3d_Builder.cxx
@@ -181,7 +181,7 @@ void ChFi3d_Builder::ExtentAnalyse()
 
 //=================================================================================================
 
-void ChFi3d_Builder::Compute()
+void ChFi3d_Builder::Compute(const Message_ProgressRange& theRange)
 {
 
 #ifdef OCCT_DEBUG // perf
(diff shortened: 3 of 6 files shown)


dpasukhi:
Dear @Artem-Viveritsa 

The CLA conditions are shared inside https://github.com/Open-Cascade-SAS/OCCT/pull/1491

many thanks for your contribution. We would be happy to resieve original bug as a Issue report and will analyze and fix it from our side.

PR #1514, opened 2026-08-29
Data Exchange - Harden malformed input handling

dpasukhi:
- Validate line lengths and preserve null-terminated text buffers;
- Parse OBJ commands with bounded views and preserve empty names;
- Validate STL and VRML records before accessing geometry data;
- Reject inconsistent STEP B-spline data using the curve construction contract;
- Resolve nested oriented edges iteratively and reject cycles;
- Add focused GTests for malformed and valid input cases.

### DIFF
--- a/src/DataExchange/TKDEOBJ/GTests/FILES.cmake
+++ b/src/DataExchange/TKDEOBJ/GTests/FILES.cmake
@@ -2,4 +2,5 @@
 set(OCCT_TKDEOBJ_GTests_FILES_LOCATION "${CMAKE_CURRENT_LIST_DIR}")
 
 set(OCCT_TKDEOBJ_GTests_FILES
+  RWObj_Reader_Test.cxx
 )
--- a/src/DataExchange/TKDEOBJ/GTests/RWObj_Reader_Test.cxx
+++ b/src/DataExchange/TKDEOBJ/GTests/RWObj_Reader_Test.cxx
@@ -1,0 +1,85 @@
+// Copyright (c) 2026 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in OCCT
+// distribution for complete text of the license and disclaimer of any warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include <RWObj_TriangulationReader.hxx>
+#include <NCollection_LinearVector.hxx>
+
+#include <gtest/gtest.h>
+
+#include <sstream>
+
+namespace
+{
+class TrackingReader : public RWObj_TriangulationReader
+{
+public:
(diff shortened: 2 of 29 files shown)


coldtobi:
Hi @dpasukhi ,
thanks for pointing me to this PR! 
 just to be sure: This PR is covering all of those CVEs? (to me it looks like it does, so thanks for confirming)

[CVE-2026-42476](https://security-tracker.debian.org/tracker/CVE-2026-42476)
[CVE-2026-42477](https://security-tracker.debian.org/tracker/CVE-2026-42477)
[CVE-2026-42478](https://security-tracker.debian.org/tracker/CVE-2026-42478)
[CVE-2026-42479](https://security-tracker.debian.org/tracker/CVE-2026-42479)
[CVE-2026-42480](https://security-tracker.debian.org/tracker/CVE-2026-42480)
[CVE-2026-42481](https://security-tracker.debian.org/tracker/CVE-2026-42481)

dpasukhi:
Dear @coldtobi 
 I used the https://gist.github.com/sgInnora/dfba083d04906283e9c92aea78e2d94a

I will recheck with your list, probably I can only tomorrow do that :(

coldtobi:
Hi @dpasukhi 

> I used the https://gist.github.com/sgInnora/dfba083d04906283e9c92aea78e2d94a

That is the same reference that Debian used, so my list is equal the list in from the gist, so if you used the same, I guess we are set. (and don't worry, there is no need for urgency)

gkv311:
@coldtobi , sorry for my intervention, but as I'm pretty far from the world of security issues, I'm curious about the process of handling these CVE reports.

It would be naive to say hat OCCT doesn't have memory handling issues, but this particular report looks very suspicious to me. Lets take a look at the item about STL reader from the gist:

> ## CVE-2026-42476 — STL ASCII Parser Heap OOB Read
> In `src/DataExchange/TKDESTL/RWStl/RWStl_Reader.cxx`, `RWStl_Reader::ReadAscii()` feeds triangle data directly into `sscanf` without first checking whether the buffer returned by `Standard_ReadLineBuffer::ReadLine()` is large enough to hold it. Craft an STL file so that `ReadLine` returns a 1–3 byte buffer and those `sscanf` calls will read past the end of the allocation.
> Two distinct code paths hit the same root cause; they share this CVE.
> **Fix**: validate `ReadLine` return length before parsing.

First strange thing in this report is referring to `sscanf` function. What is strange about it? The code in STL reader that used `sscanf` function has been [removed 9 years ago](https://github.com/Open-Cascade-SAS/OCCT/commit/07bbde451a11595140977b457e50b60c35cab354#diff-ad9a292ecd6eff686f9bf977e94647515c12f09b378d809fa9600b62c25542e4L317) since OCCT 7.3.0! The code of OCCT 7.8.1 still has leftovers defining `sscanf_l` macros, which are however used nowhere.

Then report suggests a crafted STL file, which however, doesn't trigger any memory issues in the code:
```
solid s
facet normal 0 0 0
outer loop
vertex 0 0 0
endloop
endfacet
endsolid

```

The other issues in the same reports looks suspicious as well - like somebody seeing the usage of unsafe functions without range checks in place, without analyzing the overall logic in the code that ensures ranges beforehand.

What is that? To me it looks like a report of a poor quality which either lists imaginary issues or doesn't provide clear instructions on how these could be triggered.

dpasukhi:
I also agree, but anyway we must to take an action :(

PR #1522, opened 2026-09-02
Modeling Algorithms - Fix Defeaturing failing to remove a corner round

### LINKED ISSUE #1521: Modeling Algorithms - Defeaturing cannot remove a corner round: adjacent-face extension is sized by the feature, not by the distance the faces must span
### Description

`BRepAlgoAPI_Defeaturing` fails with `BOPAlgo_AlertUnableToRemoveTheFeature` when removing a convex cylindrical corner round, while the concave fillets of the very same part are removed without trouble.

Reproducer, STEP file and the proposed patch: https://gist.github.com/GabrielJMS/6c722dc8e675c42211af777e342aeb25

**Root cause.** `FillGap::ExtendAdjacentFaces` sizes the extension of the adjacent faces from the bounding box of the *feature*:

```cpp
// BOPAlgo_RemoveFeatures.cxx
Bnd_Box aFeatureBox;
BRepBndLib::Add(myFeature, aFeatureBox);
const double anExtLength = sqrt(aFeatureBox.SquareExtent());
```

That length has nothing to do with how far the adjacent faces actually have to travel to meet each other, and for a small feature between faces meeting at a shallow angle it is too short.

Concretely, for face 32 of the attached part (a cylindrical corner round, area 56.5, z in [25, 30]) the extension is `10.7278`. The extended neighbour plane reaches x = -58.44, but it needs x ~ -56.85 to intersect the extended neighbour cylinder - **it misses by ~1.6 mm**. Measured splits of that extended face after the intersection pass in `TrimExtendedFaces`:

| extension | splits of the extended plane | kept after the ext-boundary filter |
|---|---|---|
| x1 (current) | 3 - only the two horizontal planes cut it | **0** |
| x2 | 6 - the cylinder cuts it too | 1 |

With no closing edge from the cylinder, every split of the extended face still carries a free boundary edge of the extension, the ext-edge filter discards all of them, `aLFTrimmed` ends up empty and `FillGap::TrimFace` falls into its last-resort branch:

```cpp
else if (aLFTrimmed.IsEmpty())
{
  // Use all splits, including those having the bounds of extended face
```

So the *untrimmed* extensions leak into the reconstruction that is handed to `BOPAlgo_MakerVolume`:

```
ADJ[2] plane  orig area  769.3  ->  1 image,  area 17.4        (98% of the face lost)
ADJ[3] plane  orig area  170.6  ->  4 images, incl. z 30->40.73 and z 14.27->25
ADJ[4] plane  orig area 2066.6  ->  2 images, incl. the whole untrimmed extension
```

`MakerVolume` then builds **0 solids** (no errors, one `BOPAlgo_AlertFaceBuilderUnusedEdges` warning), and `RemoveFeature` takes the `!anExpS.More()` branch, emitting `BOPAlgo_AlertUnableToRemoveTheFeature`.

On this part, reaching that fallback branch correlates perfectly with failure: 20 of the 48 faces reach it and all 20 fail, while all 9 faces that are removable today reach it zero times.

### Expected Behavior

The corner round is removed, leaving a valid 47-face solid - the same way the concave fillets of the part already are.

### Actual Behavior

```
=== removing face 3 ===
    (no message - face removed)
=== removing face 32 ===
  Warning: Unable to remove the feature
```

The result shape comes back unmodified (48 faces) and `HasWarnings()` reports `BOPAlgo_AlertUnableToRemoveTheFeature`.

```tcl
pload MODELING
stepread defeaturing_corner_round.step s *
renamevar s_1 sh
explode sh F

foreach i {3 32} {
  puts "=== removing face $i ==="
  set rc [catch {removefeatures r$i sh sh_$i} msg]
  puts "  $msg"
}
```

Faces 17, 20, 22, 28, 29 and 32 (`TopExp_Explorer` / `explode` order) are the convex corner rounds that fail; faces 3, 7, 10, 14, 23, 26, 31, 40 and 43 are concave fillets of the same part that succeed. A C++ sweep over all 48 faces is in `repro.cxx` in the gist.

### Proposed Solution

Treat the degenerate trim as the retry signal it already is: flag it in `TrimFace`, and in `FillGap::Perform` retry extend + trim with a doubled extension length (3 attempts), keeping the first attempt's result if none of them trims cleanly - so cases that pass through the fallback today are bit-for-bit unchanged.

`proposed_fix.patch` in the gist applies to current master. Measured on the attached part with the sweep in `repro.cxx`:

```
unpatched: removable:  9 / 48
patched:   removable: 15 / 48
```

The 6 newly removable faces are exactly the convex corner rounds 17, 20, 22, 28, 29 and 32; each yields a valid single solid with 47 faces and a consistent volume change. Nothing moves in the other direction - the diff of the two sweeps contains only those 6 lines.

Happy to open a PR with the patch, a `bugs/modalg` test case and the STEP file if the approach looks right.

### Operating System

Linux

### Compiler

GCC

### Bitness

64-bit

### OCCT Version

latest

### Additional Files

https://gist.github.com/GabrielJMS/6c722dc8e675c42211af777e342aeb25 - STEP file, DRAW script, C++ reproducer and the proposed patch.

Reproduced on 7.7.2, and `BOPAlgo_RemoveFeatures.cxx` is logic-identical on master (every commit touching it since `V7_7_2` is clang-format / refactor only).

GabrielJMS:
## Pre-Submission Checks

Fixes #1521.

## Problem / Motivation

`BRepAlgoAPI_Defeaturing` cannot remove a convex cylindrical corner round, while the concave fillets of the very same part are removed without trouble. The operation reports `BOPAlgo_AlertUnableToRemoveTheFeature` and returns the shape unmodified.

`FillGap::ExtendAdjacentFaces` sizes the extension of the adjacent faces from the bounding box of the *feature*:

```cpp
Bnd_Box aFeatureBox;
BRepBndLib::Add(myFeature, aFeatureBox);
const double anExtLength = sqrt(aFeatureBox.SquareExtent());
```

That length is unrelated to how far the adjacent faces actually have to travel to meet each other. For the part attached to #1521 the extension is `10.7278`, and the extended neighbour plane reaches x = -58.44 where it needs x ~ -56.85 to intersect the extended neighbour cylinder - it misses by ~1.6 mm.

With no closing edge from that intersection, every split of the extended face still carries a free boundary edge of the extension, the ext-edge filter discards all of them, `aLFTrimmed` ends up empty and `FillGap::TrimFace` falls into its last-resort branch:

```cpp
else if (aLFTrimmed.IsEmpty())
{
  // Use all splits, including those having the bounds of extended face
```

The untrimmed extensions then leak into the reconstruction handed to `BOPAlgo_MakerVolume` - one adjacent face keeps 2% of its area, another contributes pieces sticking 10.7 mm above and below the solid, a third contributes its whole extended plane. `MakerVolume` builds **0 solids** (no errors, one `BOPAlgo_AlertFaceBuilderUnusedEdges` warning) and `RemoveFeature` takes the `!anExpS.More()` branch.

On that part, reaching the fallback branch correlates perfectly with failure: 20 of the 48 faces reach it and all 20 fail, while all 9 faces removable today reach it zero times.

## Proposed Solution

Treat the degenerate trim as the retry signal it already is:

- `FillGap::TrimFace` raises a new `myTrimDegenerated` flag when it takes the "use all splits" branch.
- `FillGap::Perform` repeats extension + trim with a doubled length, up to three attempts, stopping at the first attempt that trims cleanly.
- If no attempt trims cleanly, the result of the **first** one is restored, so every case that passes through the fallback today produces exactly the result it produces now.
- `ExtendAdjacentFaces` takes the length as a parameter instead of recomputing it.

The retry only runs for features that currently reach the degenerate branch, so unaffected cases pay nothing.

### Original
image

### Defeatured
image

Built and tested on Linux x64, GCC 13.3, on this branch's base (`IR` @ `c7ea5bf865`, `8.1.0.dev1`).

`testgrid boolean removefeatures`, same build with the patch toggled:

| | result |
|---|---|
| without the patch | 1 FAILED, 27 OK, 31 SKIPPED |
| with the patch | **28 OK**, 31 SKIPPED |

The single failing case is the new test, and the per-case diff of the two runs is exactly one line:

```
< CASE boolean removefeatures G5: FAILED (error)
---
> CASE boolean removefeatures G5: OK
```

None of the 27 existing cases changes state.

Checks performed:

## CLA Confirmation

- [x] I confirm that I have read the contribution requirements, and that I have signed and submitted the CLA or I am covered by an approved company CLA.

CLA ID / submission reference: Pending

## Review Notes

**This PR needs a data file added to the test dataset.** `tests/boolean/removefeatures/G5` loads `bug1521_corner_round.step`, which is attached to #1521 (https://gist.github.com/GabrielJMS/6c722dc8e675c42211af777e342aeb25) and is not yet in `opencascade-dataset`. The test will report SKIPPED until it is added - happy to submit it wherever it should go.

31 of the 58 `removefeatures` cases were SKIPPED locally because their data files are absent from the public `opencascade-dataset-7.9.0` archive (`bug28840_*.brep`, `bug29481_L3.step`, `bug29481_window_slots.brep`, `bug30094.brep`, `bug26689_nist_ctc_01_asme1_ap242.stp`), so I could not exercise those - CI covers them.

The number of attempts (3) and the doubling factor are the conservative values that fix the reported part; happy to change either if you would rather scale the extension differently, for instance from the bounding box of the adjacent faces instead of retrying.

### DIFF
--- a/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_RemoveFeatures.cxx
+++ b/src/ModelingAlgorithms/TKBO/BOPAlgo/BOPAlgo_RemoveFeatures.cxx
@@ -52,6 +52,9 @@
 #include <TopTools_ShapeMapHasher.hxx>
 #include <NCollection_IndexedDataMap.hxx>
 
+#include <algorithm>
+#include <cmath>
+
 //=======================================================================
 // static methods declaration
 //=======================================================================
@@ -374,7 +377,8 @@ public: //! @name Constructors
   //! Empty constructor
   FillGap()
       : myRunParallel(false),
-        myHasAdjacentFaces(false)
+        myHasAdjacentFaces(false),
+        myTrimDegenerated(false)
   {
   }
 
@@ -440,17 +444,82 @@ public: //! @name Perform the operation
         return;
       }
 
-      // Extend the adjacent faces keeping the connection to the original faces
-      NCollection_IndexedDataMap<TopoDS_Shape, TopoDS_Shape, TopTools_ShapeMapHasher>
-        aFaceExtFaceMap;
-      ExtendAdjacentFaces(aMFAdjacent, aFaceExtFaceMap, aPS.Next());
-      if (!aPS.More())
+      // The extension length is derived from the size of the feature, which may be too
+      // small for the extended adjacent faces to reach each other. In that case trimming
+      // of the extended faces degenerates and the untrimmed extensions leak into the
+      // result, making the reconstruction unusable. Retry with a longer extension,
+      // keeping the result of the first attempt if none of them succeeds.
+      Bnd_Box aFeatureBox;
+      BRepBndLib::Add(myFeature, aFeatureBox);
+      const double aFeatureSize = sqrt(aFeatureBox.SquareExtent());
+
+      // The distance at which the adjacent surfaces meet is not related to the size of
+      // the feature - for faces meeting at a small angle it grows without bound - so the
+      // number of attempts is taken from the size of the solids the feature belongs to.
+      // The last attempt is then the first extension spanning the whole model; beyond
+      // that length the extended faces already cover it and doubling cannot bring them
+      // together anymore.
+      Bnd_Box aSolidsBox;
+      for (int i = 1; i <= mySolids.Extent(); ++i)
       {
-        return;
+        BRepBndLib::Add(mySolids(i), aSolidsBox);
+      }
+      const double aSolidsSize = aSolidsBox.IsVoid() ? 0.0 : sqrt(aSolidsBox.SquareExtent());
+      const int    aNbAttempts =
+        (aSolidsSize > aFeatureSize)
+             ? std::max(THE_NB_EXTENSION_ATTEMPTS,
+                     1 + static_cast<int>(std::ceil(std::log2(aSolidsSize / aFeatureSize))))
+             : THE_NB_EXTENSION_ATTEMPTS;
+
+      NCollection_IndexedDataMap<TopoDS_Shape,
+                                 NCollection_List<TopoDS_Shape>,
+                                 TopTools_ShapeMapHasher>
+                                     aFacesFirst;
+      occ::handle<BRepTools_History> aHistoryFirst;
+
+      Message_ProgressScope aPSExt(aPS.Next(2), nullptr, aNbAttempts);
+      for (int anAttempt = 0; anAttempt < aNbAttempts; ++anAttempt)
+      {
+        myTrimDegenerated = false;
+        myFaces.Clear();
+        myHistory = new BRepTools_History();
+
+        Message_ProgressScope aPSAttempt(aPSExt.Next(), nullptr, 2);
+
+        // Extend the adjacent faces keeping the connection to the original faces
+        NCollection_IndexedDataMap<TopoDS_Shape, TopoDS_Shape, TopTools_ShapeMapHasher>
+          aFaceExtFaceMap;
+        ExtendAdjacentFaces(aMFAdjacent,
+                            aFeatureSize * (1 << anAttempt),
+                            aFaceExtFaceMap,
+                            aPSAttempt.Next());
+        if (!aPSAttempt.More())
+        {
+          return;
+        }
+
+        // Trim the extended faces
+        TrimExtendedFaces(aFaceExtFaceMap, aPSAttempt.Next());
+
+        if (!myTrimDegenerated)
+        {
+          break;
+        }
+
+        if (anAttempt == 0)
+        {
+          aFacesFirst   = myFaces;
+          aHistoryFirst = myHistory;
+        }
       }
 
-      // Trim the extended faces
-      TrimExtendedFaces(aFaceExtFaceMap, aPS.Next());
+      if (myTrimDegenerated)
+      {
+        // None of the attempts produced a properly trimmed reconstruction -
+        // keep the result of the first one.
+        myFaces   = aFacesFirst;
+        myHistory = aHistoryFirst;
+      }
     }
     catch (Standard_Failure const&)
     {
@@ -576,16 +645,11 @@ private: //! @name Private methods performing the operation
   //! Extends the found adjacent faces and binds them to the original faces.
   void ExtendAdjacentFaces(
     const NCollection_IndexedMap<TopoDS_Shape, TopTools_ShapeMapHasher>& theMFAdjacent,
+    const double                                                         theExtLength,
     NCollection_IndexedDataMap<TopoDS_Shape, TopoDS_Shape, TopTools_ShapeMapHasher>&
                                  theFaceExtFaceMap,
     const Message_ProgressRange& theRange)
   {
-    // Get the extension value for the faces - half of the diagonal of bounding box of the feature
-    Bnd_Box aFeatureBox;
-    BRepBndLib::Add(myFeature, aFeatureBox);
-
-    const double anExtLength = sqrt(aFeatureBox.SquareExtent());
-
     const int             aNbFA = theMFAdjacent.Extent();
     Message_ProgressScope aPS(theRange, "Extending adjacent faces", aNbFA);
     for (int i = 1; i <= aNbFA && aPS.More(); ++i, aPS.Next())
@@ -593,7 +657,7 @@ private: //! @name Private methods performing the operation
       const TopoDS_Face& aF = TopoDS::Face(theMFAdjacent(i));
       // Extend the face
       TopoDS_Face aFExt;
-      BRepLib::ExtendFace(aF, anExtLength, true, true, true, true, aFExt);
+      BRepLib::ExtendFace(aF, theExtLength, true, true, true, true, aFExt);
       theFaceExtFaceMap.Add(aF, aFExt);
       myHistory->AddModified(aF, aFExt);
     }
@@ -835,6 +899,11 @@ private: //! @name Private methods performing the operation
     }
     else if (aLFTrimmed.IsEmpty())
     {
+      // The extended face could not be trimmed by the bounds of the original face -
+      // most likely the extension is too small for the extended faces to intersect
+      // each other. Signal it, so that the extension can be retried with a greater
+      // length.
+      myTrimDegenerated = true;
       // Use all splits, including those having the bounds of extended face
       anExpF.ReInit();
       for (; anExpF.More(); anExpF.Next())
@@ -888,7 +957,12 @@ private: //! @name Fields
   NCollection_IndexedMap<TopoDS_Shape, TopTools_ShapeMapHasher> mySolids;                //!< Solids participating in the feature removal
   NCollection_IndexedDataMap<TopoDS_Shape, NCollection_List<TopoDS_Shape>, TopTools_ShapeMapHasher> myFaces;  //!< Reconstructed adjacent faces
   occ::handle<BRepTools_History> myHistory;                //!< History of the adjacent faces reconstruction
+  bool myTrimDegenerated;                 //!< Flag to show that trimming of the extended faces has failed
   // clang-format on
+
+  //! Least number of attempts to extend the adjacent faces, each one doubling the
+  //! extension length; more are made when the model is much larger than the feature
+  static constexpr int THE_NB_EXTENSION_ATTEMPTS = 3;
 };
 
 typedef NCollection_DynamicArray<FillGap> VectorOfFillGap;
--- a/src/ModelingAlgorithms/TKBO/GTests/BRepAlgoAPI_Defeaturing_Test.cxx
+++ b/src/ModelingAlgorithms/TKBO/GTests/BRepAlgoAPI_Defeaturing_Test.cxx
@@ -1,0 +1,148 @@
+// Copyright (c) 2026 OPEN CASCADE SAS
+//
+// This file is part of Open CASCADE Technology software library.
+//
+// This library is free software; you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License version 2.1 as published
+// by the Free Software Foundation, with special exception defined in the file
+// OCCT_LGPL_EXCEPTION.txt. Consult the file LICENSE_LGPL_21.txt included in OCCT
+// distribution for complete text of the license and disclaimer of any warranty.
+//
+// Alternatively, this file may be used under the terms of Open CASCADE
+// commercial license or contractual agreement.
+
+#include "BOPTest_Utilities.pxx"
+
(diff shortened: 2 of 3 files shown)


GabrielJMS:
### Additional investigation: the same parameter is wrong in the other direction

I kept digging on a second model that Defeaturing also fails on, and it turns out to be the
*same* variable behaving badly in the opposite direction. Posting the findings here so a
maintainer can judge whether this PR should stay narrow or grow. **This PR is unchanged** —
I prototyped a generalisation and then rejected it on measurements (details at the end).

#### The parameter

```cpp
// BOPAlgo_RemoveFeatures.cxx, FillGap::ExtendAdjacentFaces
Bnd_Box aFeatureBox;
BRepBndLib::Add(myFeature, aFeatureBox);
const double anExtLength = sqrt(aFeatureBox.SquareExtent());
```

The feature's bounding-box diagonal is unrelated to the distance the adjacent faces have to
span, so it can be wrong either way:

* **too small** — the extended faces never reach each other, trimming degenerates, and the
  untrimmed extensions leak into the result. This is what this PR fixes, and it is what
  issue #1521 reports.
* **too large** — for an adjacent face much smaller than the feature, the extrapolated
  surface becomes self-interfering. On the second model, adjacent faces of 4.09, 8.77 and
  36.96 mm² are extended to 18204, 12332 and 12371 mm² — factors of 335x to 4455x. The
  feature there is a pair of long, thin fillet tori: the bbox diagonal (48.5 mm) measures
  their *length*, while the gap to be spanned is their cross-section.

#### What the over-extension does

`FillGap::TrimFace` passes the original face's non-feature edges to `aGFTrim` as both the
trimming tools and the orientation references (`aMEdgesToCheckOri`). Fed the self-interfering
extension, that general fuse **loses those argument edges outright**. `TakeModified` then
returns an empty list for them, `FindExtraShapes` has nothing to check orientation against,
the connexity-block filter discards every candidate, and the adjacent face is left
unreconstructed. `BOPAlgo_MakerVolume` gets one face instead of five and builds no solid.

The lost argument reproduces standalone, with no Defeaturing involved — feeding the 19
extended-face splits plus the 2 orientation edges straight into `BOPAlgo_Builder`:

```
arguments: 19 faces, 2 edges
errors=0  warnings=1
  15x BOPAlgo_AlertSelfInterferingShape
  75x BOPAlgo_AlertTooSmallEdge
   2x BOPAlgo_AlertAcquiredSelfIntersection
   2x BOPAlgo_AlertFaceBuilderUnusedEdges
  EDGE len=1.20519   IsDeleted=1 nModified=0   <-- LOST
      nearest of 19 argument faces: distance 9.20e-12
  EDGE len=33.2813   IsDeleted=1 nModified=0   <-- LOST
      nearest of 19 argument faces: distance 8.54e-14
```

Both edges lie on an argument face to machine precision and are discarded anyway, with
`HasErrors() == false`. Whether that is acceptable behaviour for `BOPAlgo_Builder` given
self-interfering input is a question for someone who knows that code better than I do; what
seems clearly wrong on the Defeaturing side is that `TrimFace` checks only
`aGFTrim.HasErrors()` and proceeds on a fuse that has just warned 94 times.

#### What I tried and rejected

I prototyped a direction-aware retry — detect the lost references, and scale the extension
*down* instead of up — and measured it before proposing it. It does not earn its place:

| approach | corner-round model (48 faces) | fillet model (156 faces) | grid |
|---|---|---|---|
| this PR (grow on degenerate trim) | 15 removable (9 before) | no change | 28 OK |
| + shrink on lost references | 15 | no case fixed | 28 OK |
| per-face cap `min(featSize, K*faceDiag)`, K in {0.5,1,2} | 15 | no case fixed | — |

Worse, a global shrink *regresses* working cases: sweeping single-face removals on the second
model, faces that succeed at the default length start failing at half of it. And with the
per-face cap the fillet case still fails — three of five adjacent faces do rebuild instead of
one, but their areas come out as 0.0025, 514 and 116 mm² from originals of 4.09, 8.77 and
4.09, so `MakerVolume` still builds nothing. Extrapolating a 0.06 mm x 39 mm BSpline sliver
simply has no meaningful answer.

So I have no fix for the over-extension class, and I did not want to add a detection path with
no case behind it and a measurable regression risk. This PR stays as reviewed: the narrow,
tested fix for the too-short direction.

I can supply the standalone `.brep` for the `BOPAlgo_Builder` reproducer above if that is
useful — I have held it back because it derives from a model that is not mine to publish.
Happy to open it as a separate issue if you would prefer that to a comment here.

*(CLA: still pending on my side, noted in the PR description.)*

GabrielJMS:
### Correction to my previous comment

My previous comment framed the second model's failure as the extension length being too large.
That framing does not hold up, and I would rather correct it here than leave it standing.

**What is wrong with it.** I proposed the over-extension as the cause but never showed that
correcting it fixes anything. It does not. I have since tried, on the fillet model:

| extension policy | result |
|---|---|
| global scale 0.05 .. 4.0 | no case fixed; small scales *regress* single-face removals that work at the default |
| per-face cap `min(featSize, K*faceDiagonal)`, K in {0.5, 1, 2} | no case fixed |
| per-face cap by the face's *narrow* 3d dimension, K in {1 .. 16} | no case fixed |

The middle row was also a bad experiment on my part: the adjacent faces there are strips
0.5-1.5 mm wide, so capping by their bounding-box *diagonal* still over-extends them roughly
tenfold. Redoing it against the narrow dimension changes nothing, so the conclusion stands, but
it was not earned the first time.

**What the selections actually do.** The same feature faces succeed or fail depending only on
which of them are selected together:

| selection | result |
|---|---|
| `{134}`, `{137}`, `{134,137}`, `{134,114}` | FAIL |
| `{134,137,114,117}`, `{134,137,135,136}` | FAIL |
| `{134,137,135,136,113,115}` | FAIL |
| `{134,137,135,136,114,115}` | FAIL |
| `{134,137,135,136,114,117}` | **OK** - 150 faces, valid |
| `{113,114,115,116,117}` | **OK** |
| `{113,114,115,116,117,134,137}` | FAIL |

Three 6-face selections, two failing and one working, so it is not the size of the selection.

**I cannot characterise it.** Every rule I formed has a counterexample in that table: it is not
whether an adjacent face keeps an edge outside the feature (a working selection has one with
zero), not the inner wire (two failing selections involve none), not thin adjacent faces (the
working selection still has 0.50 mm and 0.06 mm strips adjacent), not tangency (every one of
these boundaries is tangent to 0.0000 deg, working and failing alike), and not shared
`TShape`s (380 edges, 380 distinct pointers; `BRepBuilderAPI_Copy` fails identically).

**What does survive from the previous comment** is the standalone observation, which reproduces
with no Defeaturing in the picture: fed the 19 extended-face splits and the 2 orientation
edges that `FillGap::TrimFace` passes to its trimming builder, `BOPAlgo_Builder` discards both
edges - `IsDeleted=1, nModified=0` - although each lies on an argument face to within 1e-13,
and it reports `HasErrors() == false`. Downstream, `TakeModified` returns nothing for them and
`FindExtraShapes` has no orientation reference left. Separately, `TrimFace` checks only
`HasErrors()` and proceeds on a fuse that has raised 94 warnings. I no longer claim this is the
cause of the failures above - it may well be a symptom of a selection that is ill-posed to
begin with - but the lost argument looks worth a look on its own.

Model, both reproducers and the `.brep` for the general fuse case:
https://gist.github.com/GabrielJMS/b1c6e7b105ce04d0f45c49c8c5f63c2f

**None of this affects this PR**, which stays exactly as reviewed: the narrow fix for the
too-short direction, covered by `bug1521_corner_round` / test `boolean removefeatures G5`,
9 -> 15 removable faces on the reported model, `boolean removefeatures` grid 28 OK unchanged.
Happy to move the above into its own issue if you would prefer it out of this thread.

GabrielJMS:
@dpasukhi Sorry, I deleted my previous comments because I think it’s actually a completely different issue, unrelated to this bug.

I’m now trying to defeature the fillet chain `{113, 114, 115, 116, 117, 134, 137}` on the attached shape, but I haven’t been able to figure out what the problem is with this one.

So, from my point of view, I think the fix is safe to go ahead with. Concerning this other case, I haven’t given up, I’ll keep investigating and propose a new PR if I figure it out.

[piston.zip](https://github.com/user-attachments/files/31761842/piston.zip)

GabrielJMS:
> @dpasukhi Sorry, I deleted my previous comments because I think it’s actually a completely different issue, unrelated to this bug.
> 
> I’m now trying to defeature the fillet chain `{113, 114, 115, 116, 117, 134, 137}` on the attached shape, but I haven’t been able to figure out what the problem is with this one.
> 
> So, from my point of view, I think the fix is safe to go ahead with. Concerning this other case, I haven’t given up, I’ll keep investigating and propose a new PR if I figure it out.
> 
> [piston.zip](https://github.com/user-attachments/files/31761842/piston.zip)

Found a solution inspired by Woo's cell-based decomposition: https://github.com/Open-Cascade-SAS/OCCT/pull/1524

dpasukhi:
There is no delays from your side ;)

At the moment, I have no ability to validate your PR :( probably it will be opened up to next weekends or maybe I will have some time on Sunday.

Thank you for your contribution

dpasukhi:
Also if possible, it Is better to create "synthetic" Google test (s) which can be unit or integration in some way. If you want screenshots, then draw still accepted, but better with no external files

We are no longer takes draw tests with external files (to simplify future life). If it is possible somehow to "isolate" the case to some syntactic geometry/topology generated by c++ as a GTest it would be amazing.

Your brep/step files will be used during review, but they will be not be merged :( im sorry

GabrielJMS:
> Also if possible, it Is better to create "synthetic" Google test (s) which can be unit or integration in some way. If you want screenshots, then draw still accepted, but better with no external files
> 
> We are no longer takes draw tests with external files (to simplify future life). If it is possible somehow to "isolate" the case to some syntactic geometry/topology generated by c++ as a GTest it would be amazing.
> 
> Your brep/step files will be used during review, but they will be not be merged :( im sorry

Ok. Thank you! Fixed!

BarionLP:
using this branch i was able to remove multiple features that I previously had to do by hand, thanks

BarionLP:
i think there could be a bug:
If the intersection of extended faces fails outright on a retry (`BOPAlgo_AlertBuilderFailed`), `TrimExtendedFaces` returns early with `myFaces` empty and `myTrimDegenerated == false`. The loop takes that as success with zero reconstructed faces,
throwing away the (partial) result of attempt 0.

GabrielJMS:
@BarionLP Thank you for the feedback! I will investigate it!

Also, have you tried it with this branch? [https://github.com/Open-Cascade-SAS/OCCT/pull/1525](https://github.com/Open-Cascade-SAS/OCCT/pull/1525)

I’ve broken down the fixes into a series of chained PRs to address the different bugs I found and make the review easier.

Issue #1543, opened 2026-09-16
Modeling Algorithms - Boolean Cut with a helical swept tool silently removes nothing when the tool exits through a planar face (IntPatch loses the boundary vertex after seam decomposition)


polygoro:
### Description

`BRepAlgoAPI_Cut(cylinder, tool)` returns the cylinder unchanged — `IsDone()` true, no errors, no warnings, volume identical to the argument — when the tool is a `BRepOffsetAPI_MakePipeShell` sweep of a circle along a helix that starts inside the cylinder and leaves it through the top planar face. Whether it fails depends only on where the tool crosses the top face relative to the cylinder's seam. `Common` returns empty and `Fuse` returns two solids for the same arguments, i.e. the operands are treated as non-interfering.

Reproduced on OCCT 7.9.3 (native, via cadquery-ocp 7.9.3.1) and OCCT 8.0.1 (native and WebAssembly builds) with identical volumes and face counts. The two files involved are unchanged on the current `IR` (f2929057ea).

**Configuration** (C++ below): cylinder `BRepPrimAPI_MakeCylinder(8, 30)`; helix as a `Geom2d_Line(u = t, v = t·pitch/2π)` on a `Geom_CylindricalSurface(r = 8)`, pitch 2, height 15, one edge, `BRepLib::BuildCurves3d`; tool = `MakePipeShell(helix)` with `SetMode(gp_Dir(0,0,1))`, circle profile r = 0.6, `BRepBuilderAPI_RoundCorner`, `MakeSolid`, translated by DZ along Z.

| DZ | tool exits the top face at | Cut volume | faces | `IntTools_FaceFace(tube, lateral)` curves |
|---|---|---|---|---|
| 15.4 | θ = 108° | 5830.86 | 13 | 20 |
| 15.5 | θ = 90° | 5832.39 | 13 | 20 |
| 16.0 | θ = 0° | 5839.06 | 12 | 20 |
| **16.2** | θ = 324° | **6031.86 (= untouched)** | 19 | **19** |
| **16.4** | θ = 288° | **6031.86** | 18 | **18** |
| **16.6** | θ = 252° | **6031.86** | 18 | **18** |

[...]

Issue #1534, opened 2026-09-08
Boolean cut on periodic cylindrical/conical face creates disconnected wire


offbynull:
### Description

A boolean cut on a helical threaded solid produces an invalid geometry even though the 3D geometry appears continuous (wire not connected / face not closed). I noticed this issue on the latest FreeCAD release (FreeCAD 1.1.3, which uses OCCT 7.8) via the additive helix feature in its part design workbench. After some back-and-forth with an LLM, it seemed to hone in on the following: 

> Adjacent edges share the same 3D vertex, but their p-curves on the periodic root face differ by exactly 9 × 2π in U.
> 
> Example:
> U1 = -23.31059358
> U2 =  33.23807365
> ΔU = 56.54866723 = 9 × 2π.
>
> Moving the cutter axially by 0.0005 mm makes the Boolean valid.
>
> Normal adjacent UV gaps were around numerical-noise scale; this one jumped nine complete angular periods. The second bad cylindrical face showed essentially the same thing.
>
> I also checked the wire's 3D/topological vertex connectivity and did not find the equivalent giant physical gap. That combination is what led me specifically to the periodic p-curve branch mismatch, rather than merely guessing “Boolean instability.”

I've attached the brep file and FCstd: [helix missing face.zip](https://github.com/user-attachments/files/31931933/helix.missing.face.zip)  (edit: updated to include before and after brep files)

I fully admit that I'm new to OCCT and don't know enough to fully understand what's going on here. My layman's understand seems to be that a wire is produced where there are 2 consecutive edges that don't connect (the previous edge's end is wildly off from the next edge's start).

### Expected Behavior

Fully closed geometry.

### Actual Behavior

Not fully closed geometry.

### Sample Code or DRAW Tcl Script

N/A - I've attached a brep file that should highlight the issue.

### Operating System

Linux

### Compiler

Clang

### Bitness

64-bit

### OCCT Version

7.8

### Additional Files

[occt_pcurve_diag.py](https://github.com/user-attachments/files/31931583/occt_pcurve_diag.py)

Issue #1496, opened 2026-08-23
BRepAlgoAPI_Common silently returns empty (or negative) for two geometrically-identical swept solids when they are distinct operands; IsDone() is True


n3r:
## Summary

`BRepAlgoAPI_Common` computes a wrong result — usually an **empty** shape,
occasionally a shape with **negative** reported volume — when both operands
are solids built by `BRepOffsetAPI_MakePipeShell` along a path with a filleted
(circular-arc) corner, i.e. any swept solid whose wall crosses a
cylinder→torus→cylinder G1-tangent face junction. `IsDone()` reports `True`
throughout, no exception is raised, and this build exposes no `HasErrors()`
API to query. The two operands are the *same shape* (same section, path,
parameters) but come from two independent construction calls — rebuilt from
source, round-tripped through STEP, or built with a slightly different
parameter — rather than being the identical in-memory `TopoDS_Shape` object.
When the operands are genuinely the same object (or a deterministic
construction happens to hash/tolerance-match), OCCT appears to take an
internal shortcut and answers correctly; the bug surfaces once that shortcut
is not taken. We could not reproduce it by building an equivalent solid from
primitives (`BRepPrimAPI_MakeCylinder` + `BRepPrimAPI_MakeTorus` fused with
`BRepAlgoAPI_Fuse`/`BRepAlgoAPI_Cut`), so the trigger looks specific to the
B-rep produced by `BRepOffsetAPI_MakePipeShell` (seam/face parametrization),
not "any solid with a tangent junction" in the abstract.

## Environment

- OCCT via the `cadquery-ocp-novtk` Python wheel, version **7.9.3.1** (package
  version string `7.9.3.1.1`), Python 3.12.
- Confirmed on **macOS 26.6.1 arm64** (Darwin 25.6.0, this report) and on
  **Linux x86_64** (`ubuntu-latest` GitHub Actions runner), same pinned wheel
  version on both. Behavior is not identical across platforms — see
  "Additional observations".
- The reproduction below is expressed with the section/path built through
  `build123d` (a thin Python layer over `OCP`/OCCT) because
  `BRepOffsetAPI_MakePipeShell` with a filleted `BRepBuilderAPI_MakeWire`
  path is verbose to hand-roll in raw OCP and we wanted the exact B-rep that
  exhibits the bug. The actual boolean operation, volume measurement, and
  STEP round-trip in the script below are **raw `OCP` calls only**
[...]

PR #1544, opened 2026-09-16
Modeling Algorithms - Restore restriction vertices lost in IntPatch seam decomposition

polygoro:
## Pre-Submission Checks

- [x] I checked existing issues, pull requests, discussions, and forum topics for related work. (#1534 touches boolean cuts on periodic faces but is a different defect; #1496 is a silent-empty `Common` with another cause.)
- [x] I followed the contribution guidance in `.github/CONTRIBUTING.md` — with one exception stated below: the CLA is not signed yet.

**CLA:** the author has not signed the Contributor License Agreement yet. If the maintainers are inclined to merge this, the CLA will be signed before integration; the analysis and the reproduction in #1543 are usable regardless.

## Problem / Motivation

Fixes #1543.

`BRepAlgoAPI_Cut(cylinder, helicalTool)` returns the cylinder untouched — `IsDone()`, no warning, volume unchanged — when the swept tool starts inside the cylinder and leaves it through a planar end face at certain angles relative to the cylinder's seam. `Common` returns empty and `Fuse` two solids for the same operands. The user-visible impact is a wrong model that looks valid: nothing downstream can tell it from a tool that lies outside the object. Reproduced on 7.9.3 and 8.0.1; the affected code is unchanged on `IR`.

Traced with a debug build to two defects in IntPatch, both in how the face-boundary vertex of a walking line survives the seam decomposition of a periodic quadric:

1. `IntPatch_ImpPrmIntersection::Perform` puts the restriction vertices (face boundary, seam crossings) on the walking lines and *then* `DecomposeResult` splits them at the seam; the pieces are rebuilt with only their two end vertices, so the boundary vertex is lost. `GeomInt_LineConstructor` then classifies a piece straddling the face boundary by its midpoint alone and discards it whole when the midpoint is outside — including the part inside the face. (z 29.6 → 30.6 against a face bounded at z = 30 is dropped; the mirror case z 29.4 → 30.4 survives.)
2. `IntPatch_RstInt::PutVertexOnLine` checks, for closed surfaces, that the found point and the line point agree in 2D with `std::abs(U - U1) > tolUClosed`. A walking line that has just crossed the seam carries U on the far side of the period, so a correct vertex differs by 2π and is rejected although `FindParameter` located it correctly. With (1) fixed this still drops the boundary vertex whenever the boundary crossing falls within the last part-turn before a seam crossing (e.g. a 60° groove profile swept through both end faces).

## Proposed Solution

- `IntPatch_ImpPrmIntersection.cxx`: after the decomposition loop at the end of `Perform`, run `IntPatch_RstInt::PutVertexOnLine` again on every resulting walking line, with the same arguments as the existing call. (+22)
- `IntPatch_RstInt.cxx`: add `ClosedDelta(a, b, period)` (shorter way round on a closed range) and use it for the four closed-surface comparisons (pre-check and existing-vertex loop, both surfaces), with the period taken from the surface's parameter range. (+46/−11)
- `tests/bugs/modalg_8/bug_gh1543`: Draw script of the reproduction (DZ = 16.2), checking `checkshape`, one solid and the cut volume.

No API change. Formatted with the repository `.clang-format`.

## Validation

Native 8.0.1 build with the patch, C++ reproduction from #1543:

| Test | before | after |
|---|---|---|
| exit angle scan, DZ 15.0 … 16.8 step 0.1, circle profile | 12 / 19 nothing removed | 0 / 19 |
| same, 60° groove profile | failures | 0 / 19 |
| groove through both end faces, DZ −5.3 … +1 (0.05 steps near −5) | 5 × nothing removed, 2 × empty result, 3 × partial cut | all correct (937.5 mm³ removed) |
| the five failure modes catalogued for this tool family (start on face, cross both faces, start 3 mm inside, long tool, partial thread) × 2 profiles | 7 / 10 fail | 10 / 10 cut, one valid solid each |
| 8 unrelated quadric × B-spline booleans (sphere/cone/torus cut by helical tubes, arc-swept tubes, quadric × quadric) | — | volumes identical to unpatched 7.9.3 to four decimals |
| previously passing configurations | — | volume changes ≤ 0.06 mm³ (section curves now end exactly on the face boundary instead of overshooting it) |
| a downstream 26-model B-Rep regression corpus (volumes and face/edge/vertex counts) | — | unchanged |

Not run locally: the Draw test `bug_gh1543` and the full DRAW/GTest suites — no DRAWEXE was available in the environment. The Draw script mirrors the C++ reproduction step for step; please treat its first CI run as its validation, and I will adjust the expected volume tolerance if the sweep differs slightly from the C++ construction.


dpasukhi:
Dear @polygoro   Thank you for your contribution.

my apologies for CLA workflow. We are in transition stage between web-sites and CLA is not orginized.
To continue with review and integration, the CLA is required,  Please dowload the file:
[Contributor_License_Agreement_0.pdf](https://github.com/user-attachments/files/32345846/Contributor_License_Agreement_0.pdf)
After filling the PDF, contact us by email (draft started by clicking in contact form: https://occt3d.com/contact/?topic=general) and email app will be opened and you can attache filled PDF.

(the signature and filling up can be electronic, and signed and filled in electronic format).
Please be sure that all non-optional fields will be filled up. We will not use any of that information in marketing or any other unrelated needs. The information will be not shared with any 3rd-parties and stored only for legal needs.

polygoro:
I signed CLA and emailed it.
Thanks.

PR #1548, opened 2026-09-21
Modeling Data - Make TopoDS_TShape's flag word atomic

gsdali:
All of `TopoDS_TShape`'s flags live in one `myState` word, so every setter is a read-modify-write of it. As a plain `int`, two threads setting two *different* flags can read the same value and write back competing results, dropping one of the two.

**This is not a misuse of the class.** A boolean operation's result shares TShapes with its inputs, so two threads working on what look like separate shapes can be mutating one TShape, and the flags are written by ordinary healing and traversal code rather than by anything the caller drives directly.

`myState` becomes `std::atomic<int>`, read with acquire and written by a compare-exchange loop so a flag write cannot lose a concurrent one. The shape type shares the same word and is preserved by the same mechanism.

### Tests

`TopoDS_TShape_Test`:

- `ConcurrentFlagWritesDoNotLoseEachOther` gives each of three threads one flag to set and asserts all three survive, repeated 200 times. **Fails before this change.**
- `ConcurrentFlagWritesPreserveShapeType` pins that the type bits sharing the same word are not disturbed by concurrent flag traffic.


gsdali:
Configure failure fixed, force-pushed.

```
CMake Error (add_test): add_test given test NAME
  "ModelingData::TKBRep::TopoDS_TShapeTest.ConcurrentFlagWritesDoNotLoseEachOther"
  which already exists in this directory.
```

My branch predated `TopoDS_TShape_Test.cxx` arriving on master, so it **replaced** that file with my two tests and added a second `FILES.cmake` entry. That would have deleted 295 lines of your tests as well as breaking configure, which is the worse half of the problem.

Now corrected: master's file is restored intact and my two concurrency tests are appended to it, under master's existing `TopoDS_TShape_Test` suite name rather than a second near-identical one, with `<thread>` and `<vector>` added to its includes. The duplicate `FILES.cmake` entry is gone.

The diff against master is now `+57` on the test file and `+46/-15` on `TopoDS_TShape.hxx`, where before it was `-295`. Apologies for the noise.

gsdali:

PR #1549, opened 2026-09-21
Data Exchange - Make IFSelect_WorkSession's error-handling guard per instance

gsdali:
`IFSelect_WorkSession` used a file-scope flag as the sentinel that makes each of its nine error-handled operations wrap itself in a `try` exactly once:

```cpp
if (errhand) { errhand = false; try { EvalSelection(sel); } catch (...) {}
               errhand = theerrhand; return iter; }
// the real work, reached only through that recursive call
```

Because the sentinel was shared by every session in the process, clearing it on one session disabled the guard on **all the others**. A second session then took the unguarded path and a `Standard_Failure` that should have been caught and reported escaped instead.

**This is not only a threading issue.** Turning error handling off on one session did the same thing to every other session, single-threaded.

The flag was only ever a mirror of the existing per-instance `theerrhand`, written as `theerrhand = errhand = ...` and never independently, so it is removed in favour of a per-instance sentinel. No lock is needed: two sessions no longer share it.

Affects `EvalSelection`, `EvaluateComplete`, `EvaluateDispatch`, `EvaluateFile`, `EvaluateSelection`, `SendAll`, `SendSelected` and `SendSplit`.

### Tests

`IFSelect_WorkSession_Test`, the first GTest for TKXSBase:

- `ErrorHandleIsPerSession` fails before this change, where the second session's failure escapes
- `ErrorHandleOffLetsFailureEscape` pins that the guard is real rather than absent, so the first test asserts a working guard rather than a vacuous one


gsdali:
Build failure fixed, force-pushed.

`IFSelect_WorkSession.cxx:2656` still referenced the removed `errhand`:

```
error: use of undeclared identifier 'errhand'; did you mean 'srand'?
 2656 |   const bool isErrorHandlingEnabled = errhand;
```

That site had been refactored to hoist the flag into a local, so it did not match the shape of the other eight and my change missed it. It now follows the same pattern as the rest:

```cpp
const bool isErrorHandlingEnabled = theerrhand && !myInErrorHandler;
if (isErrorHandlingEnabled)
{
  myInErrorHandler = true;
```

The restore sites in that block were already converted, which is why only the two entry references failed. No bare `errhand` references remain.

gsdali:

PR #1551, opened 2026-09-21
Modeling Algorithms - Size faces from the computed density in BRepLib_PointCloudShape

gsdali:
`NbPointsByDensity` resolves `theDensity == 0` to `computeDensity()`, validates that, then sizes every face from `theDensity` **again**.

With the documented density `0` the per-face count is `ceil(anArea / 0)`, which saturates, so the six faces of a box summed to **-6 instead of 60**, and `GeneratePointsByDensity` went on to request `INT_MAX` points per face.

### Tests

`BRepLib_PointCloudShape_Test`, covering the auto-density count and its equality with the same density passed explicitly.


gsdali:
The Ubuntu failure here is **not in this change** and I cannot re-run it without repository admin rights.

It failed in the `Add NuGet sources` step, before any of OCCT compiles:

```
Cannot open assembly 'If': No such file or directory.
##[error]Process completed with exit code 2.
```

That is the vcpkg/NuGet binary-cache setup, and the same step succeeded on the other PRs I opened at the same time, so it looks like a one-off. Could someone with access re-run that job?

I have deliberately not force-pushed a no-op commit to retrigger it, since that would rewrite the SHA and move any review comment anchors for no change in content. Happy to do so if you would rather have it retriggered that way.

gsdali:

PR #1553, opened 2026-09-21
Data Exchange - Synchronise Interface_Static's parameter table

gsdali:
`Interface_Static`'s backing store is a process-global `NCollection_DataMap` keyed by parameter name, holding `MoniTool_TypedValue` objects whose string payload is mutated in place. Every STEP and IGES read and write goes through it, so **concurrent access is ordinary use rather than misuse**.

Without synchronisation a concurrent `Init` or `SetCVal` can rehash the map or rewrite a `TCollection_HAsciiString` buffer while another thread reads it. `SetCStringValue` in particular clears the shared buffer and then appends to it, so a concurrent reader can observe it **empty**. That is a memory-safety hazard, not only a stale read.

A function-local recursive mutex now covers the body of every entry point that touches the table.

**Recursive is required rather than preferred:** `Init`'s `Interface_ParamMisc` branch, its `'&'` edit branch and `Init(char)` all re-enter `Static()` on the same thread, and the very first `STEPControl_Controller` construction makes about fifty such calls, which a plain mutex self-deadlocks on.

### This is deliberately a partial fix

It does **not** claim to make concurrent data exchange correct. `Interface_Static` is used as an implicit parameter-passing channel: a caller's `SetCVal` is read many frames below inside `Transfer` or `ReadFile`, so two operations setting different values for the same parameter still cross-talk however well the accessors are locked. Closing that needs the surrounding `XSControl_Controller` ownership to change.

What this closes is the memory-safety hole in the storage itself. I would rather say that plainly than have the scope over-read.


gsdali:
MSVC build failure fixed, force-pushed.

```
Interface_Static_Test.cxx(48,27): error C3493: 'aIterations' cannot be implicitly
captured because no default capture mode has been specified
```

`aIterations` is a `constexpr int` read inside a lambda with an explicit capture list. Clang and GCC accept that; MSVC requires it captured. Added to the capture list:

```cpp
aThreads.emplace_back([aThreadIndex, aIterations, &aFailures]() {
```

I checked my other test files for the same pattern while I was here. `BSplCLib_Cache_Test` uses a default `[&, aThreadIndex]` capture so it is unaffected, and `TopoDS_TShape_Test`'s `constexpr` is only read outside its lambdas.

gsdali:

PR #1554, opened 2026-09-21
Modeling Data - Synchronise BSplCLib_Cache/BSplSLib_Cache span rebuilds

gsdali:
`BSplCLib_Cache` and `BSplSLib_Cache` each hold the span they last built. Evaluating a parameter outside that span rebuilds it **in place, from a const evaluator**, so two threads working at parameters in different spans make one cache rebuild under the other's feet, and the second reads polynomial coefficients belonging to a span its parameter is not in.

**The result is a wrong point, not only a sanitiser report.**

This is reached without any caller sharing a cache deliberately. `GeomAdaptor_Curve` and `GeomAdaptor_Surface` build one cache per adaptor and hand it to every caller of their const evaluators, and they also do a check-then-act on the cache handle itself, so the handle can be replaced while another thread is dereferencing it.

Each cache method now locks the cache's own recursive mutex around its body, and the adaptors' lazy cache construction is serialised so the check-then-act cannot tear. Recursive because a rebuild re-enters the cache's own accessors on the same thread.

### Tests

`ConcurrentEvaluationAcrossSpans` added to the existing `BSplCLib_Cache` tests. Eight threads alternate between the two ends of a three-span curve, so consecutive evaluations force rebuilds against each other, and each thread checks its own results against `BSplCLib::D0` rather than relying only on a sanitiser.

### Note on OCCT#1076

[#1076](https://github.com/Open-Cascade-SAS/OCCT/pull/1076) would rename these classes to `BSplCLib_CacheGrid`/`BSplSLib_CacheGrid` but keeps the identical unsynchronised rebuild. If that lands first I am happy to retarget this at the renamed classes rather than have it conflict.


gsdali:
Checked this against `IR` pre-emptively, because a sibling PR of mine (#1548) turned out to be **replacing** a test file `master` had gained rather than appending to it, which would have deleted 295 lines of your tests. I wanted to be sure this one is not the same shape, since `IR` also touches `BSplCLib_Cache_Test.cxx`.

**It is not.** A dry-run merge into `IR` gives exactly one conflict and no surprises:

```
CONFLICT (content): src/FoundationClasses/TKMath/GTests/BSplCLib_Cache_Test.cxx
```

- **No source-file conflict at all.** This PR touches `BSplCLib_Cache`, `BSplSLib_Cache` and the two `GeomAdaptor` pairs; `IR` touches `BSplCLib.cxx`, `BSplCLib_2.cxx`, `BSplCLib_CacheParams.hxx` and `BSplCLib_CurveComputation.pxx`. Disjoint.
- **The test conflict is a tail collision, not a replacement.** `IR` appends `NearRepeatedKnotUsesActualSpan` after the same seven tests this branch appends `ConcurrentEvaluationAcrossSpans` after. No name collision, nothing deleted; both tests should survive the resolution.

I also checked whether `IR` supersedes this, which is a mistake I have made before (carrying a patch upstream had already fixed better). It does not: `IR`'s `BSplCLib_CacheParams::IsCacheValid` change drops the next-knot epsilon check, which is span-selection correctness. This PR adds synchronisation around the rebuild. Orthogonal, and their change does not remove the race, it only makes rebuilds rarer.

I have **not** resolved it yet, on purpose: `master` has not moved since 2026-08-24, so I would rather rebase once after 8.0.2 ships and `master` absorbs `IR` than guess at the target now. Happy to resolve it sooner if that suits your integration order better.

